PR #13, opened 2019-11-26
remove pytest-runner

ben-bay:
remove pytest-runner and deprecated setup.py entries


setup.py
```diff
@@ -100,8 +100,6 @@ def extras_require():
     packages=find_packages(exclude=["tests.*", "tests"]),
     install_requires=install_requires(),
     extras_require=extras_require(),
-    setup_requires=["pytest-runner"],
-    tests_require=["pytest"],
```
lucpeterson: these aren't required anymore?

ben-bay: No, per the deprecation notice [here](https://pypi.org/project/pytest-runner/), those are now actively discouraged.

PR #15, opened 2019-11-26
Add travis auto pypi deployment

ben-bay:
added travis pypi deployment logic

fixed user name


ben-bay:
> Looks good, the user comments seem to think this needs to be done once manually to start. Not sure if that is travis related or pypi.

Good to know. Will try it out again.

PR #31, opened 2019-12-04
Fix dependencies

ben-bay:


requirements/release.txt
```diff
@@ -3,10 +3,12 @@ celery[redis]>=4.3.0
 coloredlogs
 cryptography
 importlib_resources; python_version < '3.7'
+maestrowf
```
lucpeterson: this really needs to be maestrowf>=1.1.5, which doesn't exist yet. so for now, we need to keep maestro install via GitHub develop

lucpeterson:
This line here is preventing the direct-from-github maestro being pulled into our zipped files:
https://github.com/LLNL/merlin/blob/65ae4e82dd7065671fa27666eab44568e4fd1d2e/setup.py#L77

Perhaps we should remove the -e from the maestro install?

lucpeterson:
Also this line should include the dev dependencies, so they can be installed via pip install merlin[dev]

https://github.com/LLNL/merlin/blob/65ae4e82dd7065671fa27666eab44568e4fd1d2e/setup.py#L37


I think we change it to `["mysql", "dev"]` that'll grab all of the dev dependencies as an extra

requirements/release.txt
ben-bay: Addressed. Added in commented-out `maestrowf>=1.1.5` to remind us to use it once it exists.

PR #33, opened 2019-12-04
Bugfix/koning/manifest update

koning:
Add the requirements files to MANIFEST.in so they are included in  the source distribution.


koning:
We are missing Changlelog info for 1.0.1 and 1.0.2 @ben-bay

PR #41, opened 2019-12-05
Develop -> master

ben-bay:
* Builtin workflow examples now follow template structure. Travis installs these before running tests on a workflow.
* Missing requirement `tabulate` now included.


lucpeterson:
change the version in merlin/__init__ to match 1.0.4

PR #42, opened 2019-12-05
Feature/koning/maestro git req

koning:
Fix the git version of meastrowf


koning:
My understanding is this is a difference between installing with pip and doing python setup.py install.

ben-bay:
No way to be sure this works until we deploy it, I guess.

koning:
You can create a sdist:
python setup.py sdist
then pip install from the tarball
pip install dist/merlin-1.0.5.tar.gz

PR #43, opened 2019-12-05
cryptography

ben-bay:
* Removed extraneous dependencies for cryptography
* Removed installation and dependency logic from Makefile -- let me know if you don't like that.


ben-bay:
> removing all of this from the Makefile will break a lot of our docs

Yeah on second thought I'll put this in after the pip install is working, which will allow us to simplify the docs too.

PR #46, opened 2019-12-06
Remove app config from docs

ben-bay:


lucpeterson:
a better change would be to make the examples more generic. in particular i think the redis config is useful

koning:
The redis app.yaml is effectively in the docs already.
https://merlin.readthedocs.io/en/latest/merlin_config.html#broker-redis
https://merlin.readthedocs.io/en/latest/merlin_config.html#results-backend-redis

lucpeterson:
ah; it looks like the amp_amqp.yaml is quoted directly in the docs. i suggest we keep it but change the paths

ben-bay:
> ah; it looks like the amp_amqp.yaml is quoted directly in the docs. i suggest we keep it but change the paths

@lucpeterson  thanks for the catch, do you like my recent change to address this?

PR #53, opened 2019-12-09
Dynamic docs version

ben-bay:
Our readthedocs is still showing the version as `1.0.0`. This should fix that.


docs/source/conf.py
```diff
@@ -16,17 +16,18 @@
 import sys
 sys.path.insert(0, os.path.abspath('../../..'))
 
+MERLIN_VERSION = __import__("merlin").VERSION
 
 # -- Project information -----------------------------------------------------
 
 project = u'Merlin'
-copyright = u'2018, MLSI'
+copyright = u'2019, MLSI'
```
lucpeterson: Can change to
```
from datetime import date
_year = date.today().year
copyright = f'{_year}, MLSI'
```

lucpeterson: this way we won't have to change it again in a few weeks

docs/source/conf.py
```diff
@@ -16,17 +16,18 @@
...
 author = u'MLSI'
 
 # The short X.Y version
-version = u'1.0.0'
+version = MERLIN_VERSION
 # The full version, including alpha/beta/rc tags
-release = u'1.0.0'
+release = MERLIN_VERSION
```
lucpeterson: do these have to be strings?

ben-bay: You want it to be a unicode string?

Issue #67, opened 2019-12-12
[FEATURE] Run workers on mutiple machines


koning:
A request has been made to run the workers for a given yaml spec on different machines.

The implementation will need to specify the hosts required for a given set of steps in the workers section of the spec. This can be accomplished by adding the "machines" keyword to the workers spec.  All the hosts listed in the machines section will need to have access to the OUTPUT_PATH.  Because of this constraint the OUTPUT_PATH must be a full path that can be checked for existence on all hosts listed in the machines keywords. If the machines keyword is not present, then those workers will be started on all hosts where merlin run-workers is executed.

An example is given below:
```
merlin:
  resources:
    task_server: celery
    workers:
      step1workers:
        args: -O fair --prefetch-multiplier 1 -E -l info --concurrency 36
        batch:
          type: local
        steps: step1
        machines: [host1A, hostB]
      step2workers:
        args: -O fair --prefetch-multiplier 1 -E -l info --concurrency 36
        steps: step2
        batch:
          type: local
        machines: [hostC]
```

koning:
See PR #68

koning:
Merged into develop

PR #68, opened 2019-12-12
Implement the machines keyword for the merlin workers section.

koning:
Implement the machines keyword for the merlin workers section. See issue #67 .


merlin/study/celeryadapter.py
```diff
@@ -223,6 +227,26 @@ def start_celery_workers(spec, steps, celery_args, just_return_command):
     local_queues = []
 
     for worker_name, worker_val in workers.items():
+        worker_machines = get_yaml_var(worker_val, "machines", None)
+        if worker_machines:
+            print("check machines = ", check_machines(worker_machines))
```
lucpeterson: change to logger?

koning: Luc, it is assigning workers to steps so that a user can run ```merlin run-workers <yaml>``` and it will only start the workers for that given machine.

ben-bay:
So under `merlin.resources.workers.<WORKER_NAME>` I can add the `machines` key, then a list of machine host names those workers are allowed to run on. After doing that, is there anything else I as a user need to do? Do I have to use `run-workers` from each of those machines to get this to work?

koning:
Once you add the machines then only the workers for the given set of steps on the specific machine will start. The workers must be individually started on all of the listed machines listed separately by the user.

Issue #58, opened 2019-12-10
[FEATURE] Add retry delays


ben-bay:
**Describe the solution you'd like**
Steps already have the capacity to restart. Add an optional delay time to this.

PR #76, opened 2019-12-18
countdown: retry delay

ben-bay:
Addresses issue #58.

Luckily, this is already built directly into celery: https://docs.celeryproject.org/en/latest/reference/celery.app.task.html 

Do we like `1` second as the default? Or would `0` be better?


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Development dependencies install via pip: `pip install "merlinwf[dev]"`
 - `merlin status <yaml spec>` that returns queues, number of connected
   workers and number of unused tasks in each of those queues
+- The step/run field `countdown`, which adds a delay in seconds to retries.
```
lucpeterson: I think this name should be more descriptive like `retry_delay`

ben-bay: For some reason that's the name celery uses but we can change it.

lucpeterson: or retry_countdown would be good too

lucpeterson:
we may need to pass the *args and **kwargs to the retry step

lucpeterson:
and i'm not really sure if it's supposed to be raise self.retry or just self.retry

ben-bay:
@lucpeterson implemented your suggestions from yesterday, but according to my tests this still isn't working properly. Debugging...

lucpeterson:
what test are you doing?

ben-bay:
> what test are you doing?

I'm just having a step `exit $(MERLIN_RESTART)` with `retry_delay` set to 100. When I monitor the restarts, they come a lot faster than they should.

Issue #56, opened 2019-12-10
[FEATURE] Add step timeouts/walltime


ben-bay:
**Describe the solution you'd like**
Add step timeout/walltime fields, thru Celery

AlexanderWinterLLNL:
Perhaps we should update this ticket to also reflect the need for testing from [closed ticket](https://github.com/LLNL/merlin/pull/77)

PR #77, opened 2019-12-18
walltime

ben-bay:
Partially addresses issue #56.

I still have yet to confirm this really works, but in theory it works.

Celery has a soft time limit, and a hard one. In this current implementation, I've used the hard time limit.


koning:
You can have a task that is just sleep inf and see if it is stopped.

ben-bay:
> You can have a task that is just sleep inf and see if it is stopped.

Good point. I ran this test and... it didn't work. Will need to dig deeper to find out why.

AlexanderWinterLLNL:
Let's re-examine this to build out testing.

PR #97, opened 2020-01-16
machines keyword -> develop

ben-bay:


ben-bay:
@koning just an entry about `machines` in the faq.

Still thinking about how to add a local cli test for this.

koning:
Yes that will be tricky unless travis can spawn multiple virtual hosts.

PR #110, opened 2020-01-29
Add an initial Dockerfile with an entrypoint running merlin by default.

koning:
To build this do:
docker build -t merlin .
in the merlin top-level repo directory

To run:
docker run --rm -ti merlin 
This will run merlin from within the continer, for the example function to work a directory will need to be mapped


lucpeterson:
cool! add some documentation

PR #111, opened 2020-01-29
web docs updates

ben-bay:
This doesn't update all of our missing/outdated/redundant web documentation, but it's a good first step.

Also, the new `alabaster` theme looks much nicer in my opinion.

Old:
Screen Shot 2020-01-28 at 4 27 59 PM

New:
Screen Shot 2020-01-28 at 4 26 27 PM


Makefile
```diff
@@ -75,7 +75,7 @@ install-dev: virtualenv
 
 # this only works outside the venv
 virtualenv:
-	$(PYTHON) -m venv $(VENV) --prompt $(PENV) --system-site-packages
+	$(PYTHON) -m venv $(VENV) --prompt $(PENV)
```
ben-bay: @lucpeterson you're right this would force folks to download more libraries. I removed it because on Quartz it was using the wrong version of Sphinx, which was stopping me from updating the docs.

These Makefile targets are pretty much only used by us, so whatever works best for us is fine by me.

ben-bay: For now I'll merge into `develop` with this reverted.

PR #115, opened 2020-01-30
Add a Dockerfile

koning:


koning:
They are assuming python has numpy. If you are not running in a virtualenv then python is python2 and that may not be the case.  The only python we install numpy into is python3 which is the case in the container environment.

PR #119, opened 2020-02-04
Update docs

koning:
Add commands in headers so they show up in the index.
Add missing arguments.
Add --broker to config.


docs/source/merlin_commands.rst
```diff
@@ -29,8 +29,8 @@ The Merlin can be output using the ``-v`` argument.
     (merlin3_7) $ merlin -v
 
 
-Information
------------
+Information (info)
```
lucpeterson: i like these additions
let's be more explicit "Information (`merlin info`)" etc

ben-bay: Will do

PR #123, opened 2020-02-11
Docs/koning/installation

koning:
More installation docs.
Add a server connection check to merlin info


docs/source/modules/installation.rst
```diff
@@ -1,28 +1,30 @@
 Installation
 ============
-.. admonition:: Prerequisites
```
ben-bay: Have you tried this, but keeping the admonition box? I rather like it.

koning: Not yet, but I will try it out now.

koning: Looks good, I added it.

PR #134, opened 2020-02-19
command line docs

ben-bay:
Attempts to complete the page of the web docs on command line options.

@koning is this the type of update you were hoping for?

### DIFF
--- a/docs/source/merlin_commands.rst
+++ b/docs/source/merlin_commands.rst
@@ -242,7 +242,7 @@ To send out a stop signal to some or all connected workers, use:
 
 .. code:: bash
 
-    $ merlin stop-workers [--spec <input.yaml>] [--queues <queues>] [--workers <regex>]
+    $ merlin stop-workers [--spec <input.yaml>] [--queues <queues>] [--workers <regex>] [--task-server celery]
 
 
 The default behavior will send a stop to all connected workers,
@@ -264,6 +264,8 @@ The ``--workers`` option allows you to pass in a regular expression of names of 
     # Stop all workers whose name matches this pattern, no matter the queue
     # Note the ".*" convention at the start, per regex
     $ merlin stop-workers --workers ".*@my_other_host*"
+
+The only currently available option for ``--task-server`` is celery, which is the default when this flag is excluded.
 
 .. attention::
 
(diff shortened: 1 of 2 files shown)


docs/source/merlin_commands.rst
```diff
@@ -235,47 +235,44 @@ with workers, such as via ``merlin stop-workers --workers``.
 
 .. _stop-workers:
 
-Stopping workers (stop-workers)
--------------------------------
+Stopping workers (``merlin stop-workers``)
+------------------------------------------
```
koning: This also needs --task_server

docs/source/merlin_commands.rst
```diff
@@ -302,8 +299,8 @@ If the specified directory does not exist Merlin will automatically create it.
 This will generate the example workflow at the specified location, ready to be run.
 
 
-Purging Tasks (purge)
----------------------
+Purging Tasks (``merlin purge``)
```
koning: purge also has --vars

docs/source/merlin_commands.rst
```diff
@@ -322,33 +319,44 @@ task server, run:
...
 
 .. code:: bash
 
     $ merlin purge <input.yaml> --steps step1 step2
 
 
-Debug Info
-----------
-More information can be output by increasing the logging level
+Status (``merlin status``)
```
koning: status has --task_server

docs/source/merlin_commands.rst
ben-bay: The reason I excluded task server is that the way merlin currently works, the only possible task server is celery. So I thought removing that choice from the docs might help users from getting confused.

(Any opinion on this @lucpeterson?)

docs/source/merlin_commands.rst
ben-bay: My reasoning here is same as for `stop-workers`

docs/source/merlin_commands.rst
ben-bay: Added.

docs/source/merlin_commands.rst
koning: Yes, that is a good reason.  I only mention it for completeness, if they have a  question about the arguments for ``merlin stop-workers -h`` where will they that information?

ben-bay: What I'll do for now is add a section about this, and mention that celery is the only option. We might consider deprecating this flag later on.

koning: Sounds good.

PR #135, opened 2020-02-19
Add  a check server access when `merlin info` is run.

koning:
and a port option to rabbitmq config options.


ben-bay:
Seems like a good addition to `merlin info`. Worked when I tested it too.

ben-bay:
This update worked for me when I tested it. Good addition

PR #144, opened 2020-02-26
Adding dynamic sampling workflow

jwhite242:
Adding runnable hello world based workflows

Document the dynamic sampling workflow

Fix up reserved variables table in docs


docs/source/modules/advanced_topics.rst
```diff
@@ -12,15 +12,315 @@ Advanced Topics
...
-      * 
+      * Run workflows using HPC batch schedulers
+      * Distribute workflows across multiple batch allocations and machines
+      * Spawn new workflows and choose study parameters/samples based upon already executed studies
 
 Interfacing with HPC systems
 ++++++++++++++++++++++++++++
 
+Another block is added to the merlin workflow specification when running on HPC systems,
+the `batch` block.  This block contains information about the batch scheduler system such
```
ben-bay: I think single backticks are italics, not monospace

docs/source/modules/advanced_topics.rst
```diff
@@ -12,15 +12,315 @@ Advanced Topics
...
+        run:
+           cmd: |
+             cd $(runs1.workspace)/$(MERLIN_SAMPLE_PATH)
+             $(LAUNCHER) <parallel-post-proc-script>
+           nodes: 1
+           procs: 36
+           depends: [sim-runs]
+           task_queue: post_proc_queue
+
+NOTE FOR ME TO TRY: run various post proc scripts, both with concurrent futures
```
ben-bay: sphinx comments just need: `.. ` at the line start

docs/source/modules/faker_sample.py
```diff
@@ -0,0 +1,43 @@
+"""
+Ideas to play with
+ - plotting max count or median max counts as function of samples?
+ - also keep track of unique name counts
+ - do similar for male and female names, or all at once?
+"""
+import sys
+import argparse
+
+from faker import Faker
```
ben-bay: So I ended up using the `names` module instead of `faker`. https://github.com/treyhunner/names
It's not as popular, but I just couldn't stand the way faker names were duplicated so often.

ben-bay:
Is this spec working and ready to run ?

ben-bay:
In the `merlin_variables.rst` conflict, does that formatting work within the table cell when you render the docs?

jwhite242:
Yes, I tested it with a little 6 iteration run at 100 samples each, so feel free to take it and abuse it.  

Also, good to know that names has more options than faker.  Even with the 600 samples there are multiple names that showed up more than 5 times in this one.  Though probably better for this example since i was hoping to use those counts/distributions to estimate how many unique names there are

jwhite242:
So that formatting fixes the table.  I found the published docs only showed the very last cell in that table when i went to look up all the reserved variables i could use

jwhite242:
One more comment on the workflow: you'll need to change the virtual env path in the spec variables.  The only additions it needs are faker and pandas for the various steps/samplers to run.  I'll push up a requirements.txt for it later today.

ben-bay:
Fixed a merge conflict

docs/source/modules/faker_demo.yaml
```diff
@@ -0,0 +1,113 @@
+   description:
+      name: Sample HPC specification
+      description: demo batch system and multiple worker configs for HPC workflows
+
+   env:
+     variables:
+       OUTPUT_PATH: ./name_studies
+       COLLECT: $(SPECROOT)/sample_collector.py
+       POST_PROC: $(SPECROOT)/sample_processor.py
+       PYTHON: /usr/WS2/white242/merlin_dev_2/venv_merlin_py3_7/bin/python
```
ben-bay: I'd rather avoid making users change this path manually. If their venv is already activated won't it already be pointing to the proper python?

ben-bay: Especially for cases when users pip-install merlin and don't even have a venv

jwhite242: When i ran it the first iteration works, and the sample generation step found the venv alright, but subsequent calls seemed to be hitting a different python.  Still need to figure out what's going on there as the python scripts in the post process step couldn't find pandas on iteration 2 either, even though it's installed in the system pythons as well.

PR #150, opened 2020-03-02
Feature/koning/display exceptions

koning:
Give the user some info (from merlin info) on why the servers are not connecting by printing any exceptions.


ben-bay:
So the primary change here is to show what the server errors are, if any?

PR #165, opened 2020-03-07
Change hello world sample generator to be much much faster

lucpeterson:
This uses numpy to generate the names using random array entry selections from the baseline dataset, instead of building up a list one entry at a time, which is very slow.

### DIFF
--- a/merlin/examples/workflows/hello/make_samples.py
+++ b/merlin/examples/workflows/hello/make_samples.py
@@ -1,6 +1,7 @@
 import argparse
 
 import names
+import numpy as np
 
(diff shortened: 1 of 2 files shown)


koning:
Can you add a check for the filepath  arguments existence when run?  Or print out the arguments  for people who just run it an don't read the code.

ben-bay:
Testing now...

ben-bay:
> Can you add a check for the filepath arguments existence when run? Or print out the arguments for people who just run it an don't read the code.

I can add this on another branch.

PR #168, opened 2020-03-09
merlinwf -> merlin

ben-bay:


koning:
There already was a py-merlin package in spack so there may be some issue with the name collision here.

ben-bay:
> There already was a py-merlin package in spack so there may be some issue with the name collision here.

Hmm looks like that's the same `merlin` we overwrote on pypi. It lists our repo as its homepage:
Screen Shot 2020-03-09 at 10 02 14 AM

ben-bay:
So I bet the spack people would let us use that domain too.

koning:
There are a few dependencies for the old py-merlin in spack:
    depends_on('py-merlin', type='build')
./exchanger/package.py
    depends_on('py-merlin', type='build')
./py-pythia/package.py

koning:
The spack.rst file will also need to be updated for merlinwf->merlin. If the change goes through.

adamjstewart:
Hi, I responded over email, but I just wanted to say, I'm fine with the `py-merlinwf` to `py-merlin` change in Spack. Do you happen to know if it's still possible to download the old `py-merlin` package from somewhere? Then we just need to decide on a name for the old `py-merlin` package.

P.S. I wrote the `py-merlin` package when I was adding `citcoms`, but I don't use Python 2, and `py-merlin` only supports Python 2, so I don't really care what happens to it.

koning:
I just tried to find it as well. I hadn’t realized that PyPi had removed it already. Originally I was planning to rename the package (maybe merlin-info) and create patches for py-pythia and exchanger. If we can’t find the original merlin package then we will need to remove those packages. Do you use the python2 interface?

koning:
Sorry, just  noticed you don't use the python2 interface.

koning:
Here is the PR:
https://github.com/spack/spack/compare/develop...koning:feature/koning/py-merlin?expand=1

ben-bay:
> The spack.rst file needs to change from py-merlinwf to py-merlin.

Addressed. Is that all we need?

lucpeterson:
There are other instances of merlinwf in the documentation. This is all I see on develop:
CHANGELOG.md
README.md
docs/source/getting_started.rst
docs/source/modules/installation/installation.rst
docs/source/spack.rst
merlin/examples/workflows/openfoam_wf/README.md
merlinwf.egg-info/PKG-INFO
merlinwf.egg-info/SOURCES.txt
setup.py

ben-bay:
> There are other instances of merlinwf in the documentation. This is all I see on develop:
> CHANGELOG.md
> README.md
> docs/source/getting_started.rst
> docs/source/modules/installation/installation.rst
> docs/source/spack.rst
> merlin/examples/workflows/openfoam_wf/README.md
> merlinwf.egg-info/PKG-INFO
> merlinwf.egg-info/SOURCES.txt
> setup.py

@lucpeterson squelched.

PR #172, opened 2020-03-09
Feature/koning/missing docker file

koning:


koning:
Did you already change the version? This is essentially a doc fix.

ben-bay:
In that case, it'll fit right in with the other doc fixes.

PR #174, opened 2020-03-11
Feature/koning/batch nodes all

koning:
Add an "all" option to the batch nodes argument.


ben-bay:
I added to the CHANGELOG for this.

Issue #173, opened 2020-03-10
[FEAT] zfill sample paths


lucpeterson:
## 🚀 Feature Request

**What problem is this feature looking to solve?**
The sample path files 0/0/0 etc end up not being in order

**Describe the solution you'd like**
zfill the samples directories, so that you get 00/00/00 etc

**Describe alternatives you've considered**
do nothing

**Additional context**
we'll have to make sure we catch everywhere the directories are written

ben-bay:
Added in #176

PR #176, opened 2020-03-11
Feature/luc/zfill dirs

lucpeterson:
This addresses #173 

Also fixed the bug whereby top level of directory hierarchy was always "0"

@ben-bay We will need to change the diagrams of directory structures for the hello world tutorial example, since the superfluous top layer will be gone. Also, Peter has removed the sample_index.txt file, which we don't need, so that'll need to be modded too


ben-bay:
Great, is Peter's removal of sample_index.txt already in?

lucpeterson:
@robinson96

lucpeterson:
i'd like to wait until Peter takes a look at this
please don't merge yet

ben-bay:
This works for me when I run it.

robinson96:
I don't see any concerns here.

lucpeterson:
@ben-bay  I think we're just waiting for the build to pass and then we're good to merge

PR #177, opened 2020-03-11
Applying fix to task creation bottleneck

lucpeterson:
@robinson96 's patch for unclogging task creation

### DIFF
--- a/merlin/common/tasks.py
+++ b/merlin/common/tasks.py
@@ -78,6 +78,8 @@ retry_exceptions = (
 
 LOG = logging.getLogger(__name__)
 
+STOP_COUNTDOWN = 60
+
 
 @shared_task(bind=True, autoretry_for=retry_exceptions, retry_backoff=True)
 def merlin_step(self, *args, **kwargs):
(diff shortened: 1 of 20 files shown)


merlin/common/tasks.py
```diff
@@ -414,13 +418,16 @@ def expand_tasks_with_samples(
 
     glob_path = "*/" * len(directory_sizes)
 
+    LOG.debug("creating sample_index")
     # Write a hierarchy to get the all paths string
     sample_index = create_hierarchy(
         len(samples), bundle_size=1, directory_sizes=directory_sizes, root="",
         n_digits=len(str(level_max_dirs))
-    )
```
ben-bay: We need this `)` in, it's breaking a few tests.

ben-bay:
Uh oh, it looks like the cli tests actually found something here... the 5 dry launch tests are failing. To me it looks like they're completely skipping over the steps with ``$LAUNCHER``.

ben-bay:
Once this is fixed, it would also be good to note on the CHANGELOG.

ben-bay:
On `feature_demo`, I'm getting this:

```
[2020-03-12 10:12:28: INFO] Executing step 'collect_X2.0.5' in '/g/g13/bay1/merlin/studies/feature_demo_20200312-101226/collect/X2.0.5'...
WARNING:maestrowf.interfaces.script.localscriptadapter:Execution returned an error: ls: cannot access /g/g13/bay1/merlin/studies/feature_demo_20200312-
101226/hello/X2.0.5/*//hello_world_output_*.json: No such file or directory
```

koning:
Seems like an issue with $(MERLIN_GLOB_PATH)

koning:
Actually the reason is that no hello step is being executed.

ben-bay:
> Actually the reason is that no hello step is being executed.

Right, entire steps are being skipped.

merlin/common/tasks.py
```diff
@@ -435,24 +442,38 @@ def expand_tasks_with_samples(
...
-            task_type, steps, samples, labels, sample_index, adapter_config, 0
-        )
-        sig.set(queue=steps[0].get_task_queue())
-        if self.request.is_eager:
-            sig.delay()
-        else:
-            self.add_to_chord(sig, lazy=False)
-        LOG.debug(f"merlin expansion task queued")
+        LOG.debug(f"queuing merlin expansion tasks")
+        for next_index_path, next_index in sample_index.traverse(conditional = lambda c: c.is_great_grandparent_of_leaf):
```
koning: This loop is not being executed in feature_demo.

lucpeterson: ok i found the reproducer

this fails for n_samples <= number of directories per level

lucpeterson: the problem looks to be that if there's a flat hierarchy, this doesn't trigger queueing anything off

PR #182, opened 2020-03-12
Bugfix/luc/shutdown task acknowledgement

lucpeterson:
Adds a return code to allow for shutdown of workers.
Does this via a special task that is launched in a delay, to allow the parent task to finish. The custom task is acknowledge as soon as it's pulled from the server, preventing re-queueing loop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bug that prevented an empty username for results backend and broker when using redis.
 - Bug that prevented `OUTPUT_PATH` from being an integer.
 - Bug that always had sample directory tree start with "0"
+- "Error" message whenever a non-zero return code is given
 
 ### Changed
 - Updated docs from `pip3 install merlinwf` to `pip3 install merlin`.
+- Scipt launching uses Merlin submission instead of subclassing maestro submit
 
 ## [1.4.1] [2020-03-06]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - A url keyword in the app.yaml file to override the entire broker or results backend configuration.
 - The `all` option to `batch.nodes`.
 - Auto zero-padding of sample directories, e.g. 00/00, 00/01 .. 10/10
+- `$(MERLIN_STOP_WORKERS)` exit code that shuts down all workers
 
 ### Fixed
 - Bug that prevented an empty username for results backend and broker when using redis.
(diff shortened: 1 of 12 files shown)


ben-bay:
Solid. Much better than my previous hack.

koning:
I get this output:
[2020-03-12 15:33:41,394: WARNING] Execution returned an error: 
[2020-03-12 15:33:41,394: WARNING] *** Shutting down all workers in 60 secs!

It may cause confusion that is is an error. I believe this is coming from maestro.

lucpeterson:
You're right. This is coming from the maestro localscriptadapter's submit function

lucpeterson:
@koning see the latest change, which pulls out the maestro submission, giving us more control. this also removes that "error" message

CHANGELOG.md
```diff
@@ -16,9 +16,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bug that prevented an empty username for results backend and broker when using redis.
 - Bug that prevented `OUTPUT_PATH` from being an integer.
 - Bug that always had sample directory tree start with "0"
+- "Error" message whenever a non-zero return code is given
 
 ### Changed
 - Updated docs from `pip3 install merlinwf` to `pip3 install merlin`.
+- Scipt launching uses Merlin submission instead of subclassing maestro submit
```
ben-bay: Script

koning:
Looks good, I'll create maestro issue for this. I am going to suggest that maestro look in a list of return codes for OK.

merlin/examples/workflows/feature_demo/feature_demo.yaml
```diff
@@ -109,12 +109,13 @@ study:
...
 
-    - name: shutdown all workers
+    - name: shutdown
       description: |
-          shutdown all workers
+          shutdown all workers attached to this queue
       run:
          cmd: |
-            exit $(MERLIN_STOP_WORKERS)
+            exit $(MERLIN_HARD_FAIL)
```
koning: Why do you want to shutdown only some of the workers?

merlin/common/tasks.py
```diff
@@ -137,15 +137,19 @@ def merlin_step(self, *args, **kwargs):
...
-            LOG.error(
-                f"*** Step '{step_name}' in '{step_dir}' hard failed. Quitting workflow."
-            )
             # TODO purge queues? function requires maestro_spec
             # router.purge_tasks("celery", ?, force=True)
 
-            # stop workers TODO make this more discriminatory, stopping only the relevant workers
-            shutdown = shutdown_workers.s()
-            shutdown.set(queue=step.get_task_queue())
+            # stop all workers attached to this queue
```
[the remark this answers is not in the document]
lucpeterson: done

merlin/examples/workflows/feature_demo/feature_demo.yaml
lucpeterson: hard_fail will stop all the workers in that workflow, since they all look at that queue; stop_workers stops all workers everywhere. i've put a note in there

koning: Ok, but in general HARD_FAIL will only stop the workers for queues associated with the specific failed step. Is this correct?

lucpeterson: yes

lucpeterson: as an alternative, it could shutdown all workers associated with all the queues in that workflow, but that would involve getting information that lives outside of the individual step. it might also have the side effect of bleeding into other workflows' workers via the default queue.

right now it's easiest to just stop workers on that specific queue, which i think should encourage users to be conscious of which queue they want to shut down

PR #195, opened 2020-03-27
Bugfix/peter/task creation bottleneck

robinson96:
Should address startup bottlenecks associated with large tasks. Finally got around to fixing small task creation.


ben-bay:
Thanks, looks like this passes the tests. What was the issue?

robinson96:
take a look at the last commit. I was accidentally filtering out hierarchies that didn't have great grandfather nodes

________________________________

PR #204, opened 2020-04-17
Bugfix/koning/lsb mcpu hosts

koning:
Patch for LSF number of hosts


lucpeterson:
Builds started failing here: https://travis-ci.com/github/LLNL/merlin/builds/157278190

I'm ok with this PR, but we need to track this down. Looks like a bad lib crept in?

ben-bay:
To me, it looks like the version test script I wrote is misbehaving.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to Merlin will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
```
ben-bay, line 6: Maybe my test script is looking for a `## [Unreleased]` here, it says the error is because of the CHANGELOG.

koning: If you look at the test logs , it is due to deplic complaining about the GNU (FOSS exception) license for mysql-connector-python.

ben-bay: Oooh. Wow maybe it's a good thing I added that random test a while back, never would've caught this normally

ben-bay: I cannot find where in the project we are using `mysql-connector-python`.

koning: I believe it is actually a celery dep, celery uses sqlalchemy for mysql I believe. Maybe these are for that dep? Jessica put this in so I'm not sure.

koning: Celery no longer depends on mysql, so I changed the celery dep in merlin to be merlin[redis,sqlalchemy] and removed the mysql dep. Ben can you add the newer sqlalchemy to the wheelhouse?

koning:
@ben-bay looks like i broke your test script, mysql no longer exists.

CHANGELOG.md
ben-bay: @koning you mean the RZ python wheelhouse?

koning: Yes, the current version is pretty old.

ben-bay: PR is up https://rzlc.llnl.gov/bitbucket/projects/LC/repos/python-wheelhouse/pull-requests/97/overview

PR #209, opened 2020-04-24
Bugfix/koning/batch shell

koning:


merlin/spec/specification.py
```diff
@@ -117,6 +117,7 @@ def process_spec_defaults(self):
         MerlinSpec.fill_missing_defaults(self.batch, defaults.BATCH["batch"])
 
         # fill in missing step section defaults within 'run'
+        defaults.STUDY_STEP_RUN["shell"] = self.batch["shell"]
```
ben-bay, line 120: Wondering how we had things working before without this...

In any case it looks like it'll never cause a KeyError. Looks good to me

koning: The shell wasn't set in the step here before.

PR #217, opened 2020-05-18
Feature/koning/vhost no leading slash

koning:
Review leading / from default vhost when using rabbitmq server.


merlin/config/broker.py
```diff
@@ -213,16 +219,19 @@ def get_connection_string(include_password=True):
     if broker not in BROKERS:
         raise ValueError(f"Error: {broker} is not a supported broker.")
 
-    if broker == "rabbitmq":
+    if broker == "rabbitmq" or broker == "amqps":
         return get_rabbit_connection(config_path, include_password)
```
lucpeterson, line 223: can we be more explicit here and add the conn="amqps" keyword?

koning: That is is the default, as it was before.

lucpeterson: I understand it's the default: it's mostly for self-documenting code. not important.

koning: Added.

PR #218, opened 2020-05-18
Bugfix/koning/monitor race

koning:
Reduce possible monitor race conditions.


koning:
I agree that is where monitor should go as we have discussed previously (use a spec to monitor specific queues), this was to eliminate the race until that code is ready.

PR #220, opened 2020-05-19
Bugfix/ben/expand name

### LINKED ISSUE #215: [FEAT] Allow programmatic assignment of description.name
## 🚀 Feature Request

**Describe the solution you'd like**
The ability to use this: `merlin run spec.yaml --vars STATE=Illinois ID=42`
And alter this:
```
description:
    name: $(STATE)_State$(ID)
    description: ...
```

ben-bay:
Fixes #215 

Note: must include user variable in `env` for name expansion to work from command line.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - The default rabbitmq vhost is now <user> instead of /<user>.
 
+### Fixed
+- A bug that did not change the filename of the output workspace nor of the provenance spec
+  when a user variable was included in the `description.name` field.
+
 ## [1.5.3]
 
 ### Fixed
--- a/merlin/common/tasks.py
+++ b/merlin/common/tasks.py
@@ -559,6 +559,9 @@ def queue_merlin_study(study, adapter):
     """
     Launch a chain of tasks based off of a MerlinStudy.
     """
+    import sys
+    print("DEBUG EXIT")
+    sys.exit()
     samples = study.samples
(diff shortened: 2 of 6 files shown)


lucpeterson:
I'd like to put some tests in that test this functionality. It seems like it could potentially break stuff. Particularly things like restart, query, etc.

lucpeterson:
is this failing tests?

ben-bay:
> is this failing tests?

Looks as if. I had these passing manually, so shouldn't be too tough to fix.

ptbremer:
I just tried this out and it does not yet fully work. The $(NAME) does get correctly expanded at the end but still after any sample generation script is run. 

Following the examples, I am creating the samples in the Merlin block 

```
merlin:
    samples:
        generate:
            cmd: |
                cp -r $(SPECROOT)/scripts $(MERLIN_INFO)/
                python $(SCRIPTS)/make_unit_samples.py <snip>
                python $(SCRIPTS)/make_scaled_samples.py <snip>               
                 
        file: $(MERLIN_INFO)/samples.npy
        column_labels: $(PARAMETERS)
```

But at the point this is executed $(SPECROOT) is still not correctly expanded. In fact it seems to be empty. 

/EpiCast-WF/output/_20200523-075936/merlin_info: No such file or directory

This is kind of strange since the directory is correctly created and named at the end

ben-bay:
@ptbremer good catch, looking at the cause of this now.

lucpeterson:
do we have a test for this?

ben-bay:
> do we have a test for this?

I've built this script-copying behavior into our feature_demo, so it's now being tested over and over again.

PR #221, opened 2020-05-22
Feature/koning/monitor upgrade

koning:
New monitor method:
Will check for tasks in queue(s) given a spec , similar to status.
If there are tasks in the queue(s), then it will check for workers.
If there are no workers arfter 10*(sleep) seconds, it will exit.
This last check is why this PR is a draft. I'm trying to protect against a failed workers launch.
Improvements to this may be, checking for workers on the specific allocation, this would require some allocation querying which may be pretty hacky. 
Alternatively, it can be eliminated completely and just check for an empty queue condition.
Other than that, this PR is good to go.


merlin/main.py
```diff
@@ -312,8 +319,67 @@ def process_example(args):
...
+        args.task_server, spec, args.steps, verbose=False
+    )
+
+    total_jobs = 0
+    total_consumers = 0
+    for name, jobs, consumers in queue_status:
+        total_jobs += jobs
+        total_consumers += consumers
+
+    if total_jobs and total_consumers == 0:
```
lucpeterson, line 340: total_jobs > 0 and total_consumers == 0?
took me a while to parse that. maybe too clever?

koning: This is my C optimization showing.

lucpeterson: oi my slow monkey brain

merlin/main.py
```diff
@@ -337,15 +337,15 @@ def check_merlin_status(args, spec):
...
+    if total_jobs > 0 and total_consumers == 0:
+        # Determine if any of the workers are on this allocation
         worker_names = spec.get_worker_names()
 
         # Loop until workers are detected.
         count = 0
         max_count = 10
         while count < max_count:
-            # This object will include the worker names and the host.
+            # This object will include the worker names and with the host names.
```
lucpeterson, line 348: eg worker_name@host

merlin/main.py
```diff
@@ -642,6 +708,27 @@ def setup_argparse():
...
+    monitor.add_argument(
+        "specification", type=str, help="Path to a Merlin YAML spec file"
+    )
+    monitor.add_argument(
+        "--steps",
+        nargs="+",
+        type=str,
+        dest="steps",
+        default=["all"],
+        help="The specific steps in the YAML file you want to query",
```
ben-bay, line 720: Can we update this help string to be more specific to `monitor`?

koning: Yes, I will change this.

merlin/main.py
```diff
@@ -312,8 +319,67 @@ def process_example(args):
         setup_example(args.workflow, args.path)
 
 
+def check_merlin_status(args, spec):
```
ben-bay, line 322: What do you think of moving status/monitor logic to `router.py` or elsewhere? Would that make more sense than `main.py`?

koning: The reason it is in main is because it doesn't require the router disconnect like the other router functions. If this is an aesthetic response, I don't have an issue with moving the function.

ben-bay: Yes it's aesthetic, this just seems a little different from the other logic in main.

PR #231, opened 2020-06-11
Bugfix for new celery versions

### LINKED ISSUE #203: [BUG] celery/restart bug
## 🐛 Bug Report

**Describe the bug**
Looks like this:
```
ERROR:celery.app.trace:Task merlin.common.tasks.merlin_step[31704a45-84d0-4b5a-8c71-4d9aa55cd9f2] raised unexpected: RuntimeError('Never call result.get() within a tas
k!\nSee http://docs.celeryq.org/en/latest/userguide/tasks.html#task-synchronous-subtasks\n',)
Traceback (most recent call last):
  File "/Users/bay1/merlin/venv_merlin/lib/python3.6/site-packages/celery/app/base.py", line 487, in run
    return task._orig_run(*args, **kwargs)
  File "/Users/bay1/merlin/merlin/common/tasks.py", line 130, in merlin_step
    raise RestartException
merlin.exceptions.RestartException

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/bay1/merlin/venv_merlin/lib/python3.6/site-packages/celery/app/trace.py", line 385, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/Users/bay1/merlin/venv_merlin/lib/python3.6/site-packages/celery/app/base.py", line 500, in run
    raise task.retry(exc=exc, **retry_kwargs)
  File "/Users/bay1/merlin/venv_merlin/lib/python3.6/site-packages/celery/app/task.py", line 716, in retry
    S.apply().get()
  File "/Users/bay1/merlin/venv_merlin/lib/python3.6/site-packages/celery/result.py", line 1027, in get
    assert_will_not_block()
  File "/Users/bay1/merlin/venv_merlin/lib/python3.6/site-packages/celery/result.py", line 43, in assert_will_not_block
    raise RuntimeError(E_WOULDBLOCK)
RuntimeError: Never call result.get() within a task!
See http://docs.celeryq.org/en/latest/userguide/tasks.html#task-synchronous-subtasks
```

**To Reproduce**
1. Run a single-step DAG with this as the step:

```
study:
    - name: step1
      description: step 1
      run:
         cmd: exit $(MERLIN_RESTART)
         restart: echo "restarted :)"
```

- Found in `merlin run --local` mode.
- Reproduced on local Mac and Quartz.

### LINKED ISSUE #225: [BUG] Celery v4.4.3 breaks merlin
## 🐛 Bug Report

Merlin breaks when attempting to use Celery version 4.4.3.

We have locked merlin at using Celery v4.4.2 to temporarily get around this, but we need a solution that allows us to use future versions of Celery.

koning:
Closes #203
Closes #225


koning:
I think we will need to test this on some real work workflows before merging.

ben-bay:
What kind of workflows do you have in mind? I've run this successfully on feature_demo and null_spec.

koning:
Real ones like the multi-physics simulation workflows.

ben-bay:
> Real ones like the multi-physics simulation workflows.

Ok, I don't have one set up so this may be a job for @lucpeterson

lucpeterson:
looks good to me: was able to run hydra via flux on lassen with celery 4.4.5 both distributed and locally (with the result.get() error no longer there)

koning:
OK, good the specific issue that may arise is that this task would not propagate its result to the other tasks due to the immutable flag setting. But since this task creates other tasks that do have the result propagated , it is probably not an issue.

lucpeterson:
so what's the right test? is this put to bed? is it to run like 1000 hello worlds?

koning:
That is the question, what is the correct test? I think that if you were not seeing a chord completing correctly, some tasks being executed before others, not waiting for all tasks to complete before executing a task.

PR #232, opened 2020-06-11
bugfix

ben-bay:


merlin/config/configfile.py
```diff
@@ -129,10 +129,15 @@ def get_config(path):
     return config
 
 def load_default_timeout(config):
+    seconds = 60 * 60 * 24
+    try:
+        config["celery"]
+    except KeyError:
+        config["celery"] = {"visibility_timeout_seconds": seconds}
```
koning, line 136: This can just be {}, the visibility_timeout_seconds will be handled below.

Issue #117, opened 2020-02-03
[FEAT] merlin block errors and outputs


ymubarka:
## 🚀 Feature Request

**What problem is this feature looking to solve?**
When I try to generate samples from the merlin block I don't get an error telling me what went wrong. It gives an error that is unrelated to the actual bug which is in the script itself. That error is suppressed when it should be outputted somewhere.

**Describe the solution you'd like**
Printing the error just like in the study block and having an out and err files.

**Describe alternatives you've considered**
Manually going every step and trying to figure out what went wrong. This takes way too long and it's not always clear where the problem is unless everything is broken down.

ben-bay:
@lucpeterson if I add a place for the stdout and stderr of the samples, do you think `merlin_info/` would be a good spot?

ben-bay:
Added with #234

PR #234, opened 2020-06-11
Log sample generation to merlin_info

ben-bay:
Addresses #117

### DIFF
--- a/merlin/study/study.py
+++ b/merlin/study/study.py
@@ -418,7 +418,14 @@ class MerlinStudy:
                     "cmd"
                 ]
                 LOG.info("Generating samples...")
-                subprocess.call(sample_generate, shell=True)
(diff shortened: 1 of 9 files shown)


merlin/study/study.py
```diff
@@ -418,7 +418,19 @@ def generate_samples(self):
                     "cmd"
                 ]
                 LOG.info("Generating samples...")
-                subprocess.call(sample_generate, shell=True)
+                sample_process = subprocess.Popen(
+                    sample_generate,
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.PIPE,
+                    shell=True,
```
koning, line 425: Ben, have you tested this in one of the parallel workflows?

ben-bay: I hadn't, looking at this now.

ben-bay: Everything in `slurm_par` appears to work with this branch.

koning: Thanks for checking.

PR #239, opened 2020-06-29
use regex to find variable tokens

ben-bay:


merlin/spec/expansion.py
```diff
@@ -133,14 +134,14 @@ def determine_user_variables(*user_var_dicts):
...
-        if "$(" in new_val:  # change to re
+        if contains_token(new_val):
             for determined_key in determined_results.keys():
                 var_determined_key = var_ref(determined_key)
                 if var_determined_key in new_val:
                     new_val = new_val.replace(
                         var_determined_key, determined_results[determined_key]
                     )
-        if "$" in new_val:  # change to re
+        if "$" in new_val:
```
koning, line 144: Maybe change this one to contains_shell or some such, "${}"

koning: and make a funtion like contains_token but for shell variables.

PR #240, opened 2020-06-29
More provenance specs + logic improvements

ben-bay:
Inside `merlin_info` we now have `<name>.expanded.yaml`, `<name>.orig.yaml`, and `<name>.partial.yaml`. The latter has only expanded its user variables section.

I would appreciate a little extra testing for this one.


merlin/study/study.py
```diff
@@ -116,33 +117,44 @@ def __init__(
         self.dag = None
         self.load_dag()
 
+    @cached_property
+    def provenance_original_spec(self):
+        return os.path.join(self.info, "orig.yaml")
```
koning, line 122: Maybe prepend orig_ to the filename. orig_filename

merlin/study/study.py
```diff
@@ -353,8 +363,9 @@ def expanded_spec(self):
         )
 
         # expand provenance spec filename
-        if "$(" in self.spec.name:
-            expanded_name = result.description["name"].replace(" ", "_") + ".yaml"
+        if "$(" in self.original_spec.name:
+            # expanded_name = result.description["name"].replace(" ", "_") + ".yaml"
+            expanded_name = "expanded.yaml"
```
koning, line 368: We should preserve the original filename.

merlin/study/study.py
```diff
@@ -161,16 +173,16 @@ def write_expanded_spec(self, dest):
...
-        full_spec = dump_with_overrides(self.spec, self.override_vars)
+        full_spec = dump_with_overrides(self.original_spec, self.override_vars)
 
         with open(dest, "w") as dumped_file:
             dumped_file.write(full_spec)
 
-        # update spec so that user_vars update will be accurate
-        self.spec = MerlinSpec.load_specification(dest)
+        # load spec so that user_vars update will be accurate
+        new_spec = MerlinSpec.load_specification(dest)
```
koning, line 182: Should we save this spec?

merlin/study/study.py
lucpeterson: yeah i think this is a good idea. which do we prefer?

prepend (orig_name.yaml)
or insert near the end (name_orig.yaml)

koning: Whichever is easier.

merlin/study/study.py
ben-bay: No, we're just using it two lines down.

ben-bay:
@koning @lucpeterson okay I have fixed the bugs that were here previously, and all tests are passing. The `Study` class is big and is starting to show some bad smells, so at some point I'd like to refactor the whole thing. But for now this appears to deliver the functionality we need.

PR #245, opened 2020-07-07
expose celery

ben-bay:
This is my first try at exposing celery configuration settings for curious merlin users.

With highest taking precedence, the config levels are:
1) default
2) app.yaml override
3) essential


lucpeterson:
I think we need to add some unit tests for the override

lucpeterson:
Have you compared the broker strings before and after the mods? perhaps something is getting munged...

ben-bay:
Weird, it passed in python 3.7 and 3.8 but not 3.6...

lucpeterson:
yeah it looks like the venv build for the py36 version failed on connecting to pypi. i'm not sure this falls under something we did

ben-bay:
I re-ran and it worked. Running broker string tests rn...

ben-bay:
@lucpeterson question: why have we been updating the `broker` and `backend` in our `app.conf` if fields of these names don't exist here? https://docs.celeryproject.org/en/stable/userguide/configuration.html

lucpeterson:
we construct the broker_url and backend_url from the settings in the broker and backend sections

the broker_url for instance looks like this: transport://userid:password@hostname:port/virtual_host

but we allow people to set the userid, password file, hostname, port, and virtual host as separate fields

merlin/celery.py
```diff
@@ -54,65 +54,46 @@
...
 
+app = Celery("merlin")
 
-app = Celery(
-    "merlin",
+# load merlin config defaults
+app.conf.update(**celeryconfig.DICT)
+
+# load config overrides from app.yaml
+if (
```
koning, line 77: Do we need to say anything if there is no override? How about just else part.

merlin/config/celeryconfig.py
```diff
@@ -0,0 +1,29 @@
+"""
+Default celery configuration for merlin
+"""
+
+from merlin.log_formatter import FORMATS
+
+
+DICT = {
+    "task_serializer": "pickle",
```
koning, line 9: We can't change these currently or the system will break. Should we expose this?

ben-bay: This isn't what's exposed; these are our defaults. However, with this PR technically everything is exposed, so users could break stuff in more ways than one.

merlin/celery.py
ben-bay: else pass? Sure, I can switch the no override log message to debug mode.

koning: I just mean change the logic of the if to only print if there is an override.

ben-bay:
@lucpeterson that makes sense, but I don't see any `backend_url` field in that above link to docs. We've been setting `app.conf.update(broker=BROKER_URI, backend=RESULTS_BACKEND_URI, ...)` but `broker` and `backend` aren't in that configuration and defaults doc (`broker_url` is, but we're not using that). Maybe I'm missing something, but I'm confused.

tests/test_celery.py
```diff
@@ -0,0 +1,23 @@
+import re
+
+from merlin.celery import app
+
+
+def test_broker_url():
```
koning, line 6: This is not going to work in general, on LANL this would fail because it is not ampqs.

koning: You can also use redis for the broker.

tests/test_celery.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    Ensure the celery application 'broker_url' roughly matches the required pattern.
+    """
+    if app.conf.broker_url:
+        assert re.match(
+            r"amqps:\/\/\w+:.+@jackalope\.llnl\.gov:\d+\/\w+", app.conf.broker_url
+        )
+
+
+def test_result_backend():
```
koning, line 16: Having a redis backend is not required. I would eliminate these tests. There ias a bug anyone, using broker_url instead of backend.

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to Merlin will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- The ability to override any value of the celery configuration thru `app.yaml` in `celery.override`.
+- 2 unit tests for the format of broker url and result backend.
```
koning, line 11: Remove this also?

merlin/data/celery/app.yaml
```diff
@@ -1,5 +1,8 @@
 celery:
```
koning, line 1: Has this been tested without the celery block?

ben-bay: Yep, with no celery block this works too.

ben-bay:
@lucpeterson this image shows the results of my latest commit-- I've added an info level print to show all celery configs being overridden. The `broker_url` appears to be successfully overridden here. It also raises an error when attempting to override a nonexistent celery config.

Screen Shot 2020-07-22 at 12 40 53 PM

Issue #201, opened 2020-04-01
[BUG] Rogue string expansion


ben-bay:
## 🐛 Bug Report

**Describe the bug**
The string "$_" expands to a path in the venv.

**To Reproduce**
1. Add "$_" anywhere in a spec
2. `merlin run <spec.yaml>`
3. Look in the provenance file.

**Expected behavior**
The string "$_" should not change.

EDIT: According to [this](https://stackoverflow.com/questions/5163144/what-are-the-special-dollar-sign-shell-variables), "$_" is a special shell variable.

lucpeterson:
I think we need to remove the expanduser and expandvars references by Merlin outside of the workers

https://github.com/LLNL/merlin/blob/54166e0920eb2aac36ba9fbd524ddd9680b2b1d1/merlin/spec/expansion.py#L83

I propose we leave user and environment variables alone and let the workers take care of them as they would at run time

ben-bay:
@lucpeterson okay, so even symbols like `~` won't work then

lucpeterson:
we should check whether having "~" in a spec file would work or not

lucpeterson:
i bring this up to address Timo's desire to use environment variables in spec descriptions for extra flexibility

ben-bay:
I see, good point. It's an easy fix, I'll run some tests to make sure what we think is happening is happening.

ben-bay:
Ohhhh I see the problem. Has nothing to do with `$`. We should leave env vars alone, to instead be interpreted by each step's shell and not meddled with by merlin.

ben-bay:
#247

koning:
To be clear, you should be able to use env variables in all sections except the steps. These are definitely used in the batch section.

lucpeterson:
Do we have to explicitly replace them or can they be replaced by the native shell? Can we center around an example?

koning:
Well the point is to replace them immediately, like paths to flux in the batch section. There are not part of a step so they won't be replaced by the env when they are run. Also in variables in the env section, those should be replaced at run, if a user wants the variable to remain unchanged , it should be in a step.

koning:
I think this is semantics, by replaced I mean made available at run with the run env variables. So yes, as long as they are in the shell when run it is fine.

lucpeterson:
[...]

PR #247, opened 2020-07-07
env variable bugfix

ben-bay:
Resolves issue #201

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Support and faq entry for `pgen` with `merlin run --pgen` and optional `--parg`.
 - Documentation on `level_max_dirs`.
+- Easier-to-read provenance specs.
+- Documentation on the new 3 types of provenance spec.
+
+### Fixed
+- Fixed Docker bug pertaining to Ubuntu.
 
 ### Fixed
 - Removed expansion of env variables in shell sections (`cmd` and `restart`) of provenance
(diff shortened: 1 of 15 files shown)


lucpeterson:
build is failing on unit test for $PATH
>       assert not TestMerlinStudy.file_contains_string(
536            self.study.expanded_spec.path, "$PATH"
537        )

ben-bay:
> build is failing on unit test for $PATH
> 
> > ```
> >   assert not TestMerlinStudy.file_contains_string(
> > ```
> 
> 536 self.study.expanded_spec.path, "$PATH"
> 537 )

Good catch. I'll reverse the test, so that env variable _should_ be found in the step's `run.cmd`.

ben-bay:
@lucpeterson @koning in the past, environment variable expansion was section-agnostic; it occurred equally on every line of the spec. I'll now be fixing this to only happen in non-shell cmd sections. Since this requires a bit of refactoring, I marked this PR as a WIP, probably to be finished by tomorrow.

lucpeterson:
So I think a lot of these changes need doc strings, comments, etc.
We also should add user documentation somewhere about when/how environment variables are applied. This is a big change and we want to be clear to users (and future us!) about what's going on.

koning:
Have you tested this on LC with any of the flux test specs?

ben-bay:
> Have you tested this on LC with any of the flux test specs?

@koning Ran `flux_par` successfully, the steps `data` and `stop_workers` didn't have a `MERLIN_FINISHED` file but there was no discernible error anywhere.

ben-bay:
> Have you tested this on LC with any of the flux test specs?

Yes, all variable expansion works as expected, however some of the flux examples are breaking, probably because the flux team has changed the interface again.

lucpeterson:
Does this work with $SYS_TYPE in the flux path definition?

lucpeterson:
specifically on different machines

ben-bay:
@lucpeterson Joe actually asked this same question. Yep $SYS_TYPE works. (aslo it's ignored in shell commands)

PR #252, opened 2020-07-14
pretty yaml

ben-bay:
Customized `MerlinSpec.dump()` to not rely on `yaml.dump`, which was making provenance specs ugly.

Before:
Screen Shot 2020-07-15 at 11 03 01 AM

After:
Screen Shot 2020-07-15 at 11 03 29 AM


lucpeterson:
Can you post example before-after output?

ben-bay:
> Can you post example before-after output?

Added at the top. We can also mess around with indentation and stuff, this is just what I came out with first.

ben-bay:
Verified that `spec.dump() == spec.dump(spec.dump())`.

PR #253, opened 2020-07-15
flux examples fix

koning:


ben-bay:
> Log this fix in the CHANGELOG, specifically what is now working that was previously broken

PR #254, opened 2020-07-16
Fix Dockerfile

fixdocker:
Hello, I represent a research group investigating build failures in
Dockerfiles. We executed a Dockerfile in your project and found that
it is not building. We are sending this Pull Request with a suggested
repair. The repair operation is based on the following URL:

- https://stackoverflow.com/a/61564831

Please note that we are using an anonymous GitHub account because of
double-blind restrictions in the submission process of conferences.


koning:
Thanks for the PR.

PR #267, opened 2020-08-04
Refactor/ben/spec logic

ben-bay:
* a test that would've alerted us of the bug in 1.7.0
* a more thorough fix of the recent bug (introduced in 1.7.0 and patched in 1.7.2 and 1.7.3)
* a few other internal spec logic improvements (turns out I'm better at writing code now than I was 2 years ago)

### DIFF
--- a/merlin/examples/dev_workflows/minimum_format.yaml
+++ b/merlin/examples/dev_workflows/minimum_format.yaml
@@ -1,0 +1,9 @@
+description:
+    name: minimum_format
+    description: a spec that should not break
+
+study:
+    - name: step1
+      description: step 1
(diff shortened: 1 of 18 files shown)


merlin/examples/dev_workflows/minimum_format.yaml
```diff
@@ -0,0 +1,9 @@
+description:
```
koning, line 1: This is good to test the other sections, but we should make sure merlin fails correctly if there is no description or study sections.

ben-bay: Added

Issue #202, opened 2020-04-09
[BUG] incorrect warning


ben-bay:
## 🐛 Bug Report

**Describe the bug**
The Maestro warning `Cannot set the submission time of 'sim_runs' because it has already been set.` appears when there should be no warning.

**To Reproduce**
1. Have a study step with a restart section
2. Go into the restart cmd using `exit $(MERLIN_RESTART)`
3. See warning in logs

**Expected behavior**

ben-bay:
@ymubarka is this issue more or less what you ran into?

ben-bay:
Fixed with #269

PR #269, opened 2020-08-05
Feature/koning/maestro remove submit message

koning:
Fix for #202


requirements/dev.txt
```diff
@@ -1,9 +1,9 @@
...
-dep-license
+#dep-license
 flake8
 isort
 pytest
 twine
-sphinx>=2.0.0
-alabaster
+#sphinx>=2.0.0
+#alabaster
```
ben-bay, line 9: What's up with these dep changes?

koning: I informed you about the issues with venv make previously.  That commit was a mistake. I'll take it out.

PR #282, opened 2020-11-14
reverted celery version

ben-bay:


koning:
It's too bad celery removes functionality like this, but the only files you need to change are the release.txt and CHANGLEOG.md files. The worker commands are still correct with the -A merlin first.

PR #285, opened 2020-12-15
Refactor/ben/spellbook for examples

ben-bay:
`spellbook make-samples` appears to work across the board for our existing examples. Other scripts still need a bit of work

@lucpeterson would you rather `spellbook collect` know about all data formats including conduit, or are you feeling a separate `spellbook conduit-collect`?

@ymubarka minor tweak to optimization.yaml to allow distributed runs.


merlin/examples/workflows/feature_demo/feature_demo.yaml
```diff
@@ -56,22 +53,22 @@ study:
       description: |
          train a learner on the results
       run:
-         cmd: python3 $(LEARN) -infile $(translate.workspace)/results.npz
+         cmd: spellbook learn -infile $(translate.workspace)/results.npz -outfile regressor.pkl
```
lucpeterson, line 56: feature_demo needs its requirements.txt updated to include the spellbook

tests/integration/run_tests.py
```diff
@@ -546,12 +546,6 @@ def define_tests():
             f"{run} {demo} --vars N_SAMPLES=2 OUTPUT_PATH=./{OUTPUT_DIR} --local",
             [
                 ReturnCodeCond(),
-                ProvenanceCond(
```
lucpeterson, line 549: why is this test removed?

lucpeterson (changes requested):
Looks like a number of the tests fail

merlin/examples/workflows/feature_demo/feature_demo.yaml
ben-bay: Good catch

tests/integration/run_tests.py
ben-bay: It was searching for a string about the predict script, which was removed to make way for spellbook predict

ymubarka (approved):
Are these like pep8 things?

ben-bay:
> Are these like pep8 things?

Some but I also added `$(RUN_TYPE)` to the optimization wf so it can be distributed

PR #286, opened 2020-12-16
Library/Req Changes

lucpeterson:
- Pin to updated celery with fix to queue purging
- Add `johnnydep` to dev reqs, a package that lets us find and list all dependencies
  - New `make reqlist` target


koning (approved):
Looks like they updated celery three versions at once, 5.0.3-> 5.0.5

ben-bay (approved):

PR #297, opened 2021-01-11
Bugfix/ben/restart error

ben-bay:


merlin/common/tasks.py
```diff
@@ -113,11 +113,19 @@ def merlin_step(self, *args, **kwargs):
         elif result == ReturnCode.DRY_OK:
             LOG.info(f"Dry-ran step '{step_name}' in '{step_dir}'.")
         elif result == ReturnCode.RESTART:
-            LOG.info(f"** Restarting step '{step_name}' in '{step_dir}'.")
+            # LOG.info(f"*** Restarting step '{step_name}' in '{step_dir}'.")
             step.restart = True
-            raise RestartException
+            if self.request.retries < self.max_retries:
```
lucpeterson, line 118: I wonder if we should try this approach instead
```
try:
    # log message about retrying
    self.retry()
except MaxRetriesExceededError:
    # another log message about exceeding max retries
    result = ReturnCode.SOFT_FAIL

File: merlin/common/tasks.py [hunk 2]
```diff
@@ -113,11 +113,19 @@ def merlin_step(self, *args, **kwargs):
...
+                self.retry()
+            else:
+                LOG.warning(
+                    f"*** Step '{step_name}' in '{step_dir}' exited with a MERLIN_RESTART command, but has already reached its retry limit ({self.max_retries}). Continuing with workflow."
+                )
         elif result == ReturnCode.RETRY:
-            LOG.warning(f"** Retrying step '{step_name}' in '{step_dir}'.")
+            LOG.warning(f"*** Retrying step '{step_name}' in '{step_dir}'.")
             step.restart = False
             raise RetryException
```
lucpeterson, line 130: This result should do the same thing as above

merlin/examples/dev_workflows/restart.yaml
```diff
@@ -0,0 +1,46 @@
...
+  run:
+    cmd: |
+      echo "sample: $(V1)"
+      exit $(MERLIN_RESTART)
+    max_retries: 3
+
+- name: indiv_followup_step
+  description: step that can start immedeatly after a sample has finished
+  run:
+      cmd: echo "followup complete :)"
```
lucpeterson, line 25: you may need to put a reference to $(V1) in here

merlin/examples/dev_workflows/restart.yaml
```diff
@@ -0,0 +1,46 @@
...
+      depends: [restart_and_sample_step]
+
+
+- name: bottleneck_step
+  description: step that can only start upon complete of all samples in previous step
+  run:
+      cmd: echo "bottleneck complete :)"
+      depends: [restart_and_sample_step_*]
```
lucpeterson, line 33: should we add a step that happens after the indiv_followup all complete? This will help us test chords of chains

tests/integration/run_tests.py
```diff
@@ -497,8 +497,15 @@ def define_tests():
             ReturnCodeCond(),
             "local",
         ),
+        "local restart": (
+            f"{run} {dev_examples}/restart.yaml --local --vars OUTPUT_PATH=./{OUTPUT_DIR}",
```
lucpeterson, line 501: i think we need a test for all the bottlenecks to complete, no?

lucpeterson (changes requested):

merlin/examples/dev_workflows/restart.yaml
ben-bay: Something like this?
```
- name: indiv_followup_step_2
  description: step that can start immediately after a followup has finished
  run:
      cmd: echo "$(V1) followup 2 complete :)"
      depends: [indiv_followup_step]
```

ben-bay: Maybe I don't understand how to make a chord. I thought it was just having `_*` after a dependency.

PR #299, opened 2021-02-25
Retry Delay

lucpeterson:
Addresses #58 #75 #76 

Add the ``retry_delay`` field to specify a step's delay in seconds when calling a MERLIN_RESTART or MERLIN_RETRY
This enables long-running batch jobs that end early to avoid grabbing the task right away.

New example ``merlin example restart_delay`` shows how to use it. Docs updated as well.

Note: does not work in --local mode. Unclear how to enable this (if possible), since local implies eager, which means tasks are run right away. ``retry_delay`` holds the task on the server for the specified number of seconds.


ben-bay (approved):
Maybe the reason my attempt at this failed was because I only ever tested it in `--local` mode. I would put a note in the docs that `--local` doesn't delay

lucpeterson:
ok i'll update some docs and push up changes

koning (approved):

PR #302, opened 2021-05-27
Remove --mpi=none default arg for srun launch.

koning:
Remove the --mpi=none run default launch argument. This can be added by
setting the launch_args argument in the batch section in the spec.


lucpeterson (approved):

koning:
Not sure why the feature_demo test fails. It works fine in the venv.

lucpeterson:
Alexander has fixed this test in his branch that is transitioning over to GitHub actions from travis

ben-bay (approved):

PR #304, opened 2021-06-07
task priority

ben-bay:


merlin/study/step.py
```diff
@@ -135,17 +136,17 @@ def clone_changing_workspace_and_cmd(
...
         """ given a maestro step dict, get the task queue"""
+        queue = str(QUEUE_TAG)
         with suppress(TypeError, KeyError):
-            queue = step_dict["run"]["task_queue"]
-            if queue is None or queue.lower() == "none":
-                queue = "merlin"
-            return queue
-        return "merlin"
+            queue += step_dict["run"]["task_queue"]
+            if queue is None or queue.lower() == f"{QUEUE_TAG}none":
```
lucpeterson, line 147: Here, queue will never be 'None'. Best to check step_dict["run"]["task_queue"] first to see if that's None

merlin/utils.py
```diff
@@ -57,6 +57,27 @@
 DEFAULT_FLUX_VERSION = "0.13"
 
 
+def get_priority(api="rabbit", priority="low"):
+    apis = ["rabbit", "redis"]
```
lucpeterson, line 61: this should be replaced with a call to see what the current broker is. we do this in a few places, eg via

broker = CONFIG.broker.name.lower()

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   restart or retry.
 - New merlin example `restart_delay` that demonstrates usage of this feature.
 
+### Fixed
+- Bug that caused the merlin server to crash if given too many tasks.
```
lucpeterson, line 15: This isn't really a bug per se. I'd say this is a "change" to enable queue message priorities, which prioritizes "real work" over generating more work. This change enables improved scalability and server stability.

lucpeterson (changes requested):

merlin/utils.py
```diff
@@ -57,6 +57,27 @@
...
+    if api not in apis:
+        LOG.error(f"Unrecognized api '{api}'! Options: {apis}")
+    priorities = ["high", "mid", "low"]
+    if priority not in priorities:
+        LOG.error(f"Unrecognized priority '{priority}'! Options: {priorities}")
+    if priority == "mid":
```
AlexanderWinterLLNL: This works, but I'd probably handle this as an enum to avoid this kind of conditional sprawl, especially since we're handling a return code, and especially if there's a possibility this gets expanded in the future.

merlin/utils.py
```diff
@@ -57,6 +57,27 @@
...
+        return 5
+    if api == "rabbit":
+        if priority == "low":
```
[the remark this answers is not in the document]
ben-bay: Yeah they go backwards for some insane reason.

merlin/utils.py
ben-bay: Would you have the input parameter be a string or an enum?

merlin/utils.py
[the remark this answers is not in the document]
lucpeterson: In case you're wondering -why- it does this, its because redis doesn't have true message priorities. Instead, it splits each queue into n distinct queues and workers go through them in alphabetical order, so 0 comes first and 9 last. We also need to modify app.conf.broker_transport_options (see https://docs.celeryproject.org/en/latest/userguide/routing.html#redis-message-priorities)

merlin/utils.py
ben-bay: Fixed it

merlin/config/utils.py
```diff
@@ -0,0 +1,20 @@
...
+    if broker == "rabbitmq":
+        if priority == "low":
+            return 1
+        if priority == "high":
+            return 10
+    if broker == "redis":
+        if priority == "low":
+            return 10
+        if priority == "high":
+            return 1
```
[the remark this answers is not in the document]
ben-bay: Added

merlin/celery.py
```diff
@@ -81,8 +82,8 @@
 
 # set task priority defaults to prioritize workflow tasks over task-expansion tasks
 task_priority_defaults = {
-    "task_queue_max_priority": get_priority(priority="high"),
-    "task_default_priority": get_priority(priority="mid"),
+    "task_queue_max_priority": get_priority("high"),
+    "task_default_priority": get_priority("mid"),
 }
 app.conf.update(**task_priority_defaults)
```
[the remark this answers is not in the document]
ben-bay: Does recent commit look okay?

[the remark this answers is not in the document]
lucpeterson: All checks failing

ben-bay:
> Also maybe need to modify here:
> 
> https://github.com/LLNL/merlin/blob/06f2e18018cdb79b19ca3229da06e1dbbeb85cc8/merlin/spec/specification.py#L310

Hm unsure what commit you're looking at but I'm already prepending the queue tag on that line.

merlin/study/step.py
```diff
@@ -135,17 +136,19 @@ def clone_changing_workspace_and_cmd(
 
     def get_task_queue(self):
         """ Retrieve the task queue for the Step."""
-        return self.get_task_queue_from_dict(self.mstep.step.__dict__)
+        return QUEUE_TAG + self.get_task_queue_from_dict(self.mstep.step.__dict__)
```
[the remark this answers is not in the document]
ben-bay: You're right, it was. Removing this hasn't fixed everything yet.

merlin/study/step.py
```diff
@@ -43,6 +43,7 @@
 
 
 LOG = logging.getLogger(__name__)
+QUEUE_TAG = "[merlin]_"
```
AlexanderWinterLLNL, line 46: Luc explained the purpose of this to me - I think it might be a good idea if this is dynamically linked to a version number, so like [merlin_1.8], for two reasons: if we need to come back and modify this again in the future, it would be ideal if we don't collide with the same problem again and have a built-in mechanism to not be tied to ; might just be good from a backwards compatibility perspective to decouple. Version number may not be the best thing to hash/distinguish on, but worth some thought.

ben-bay: Interesting idea. One problem with adding the version would be that users' old queues wouldn't work after an update. This update already will basically throw out old task queues, but that's necessary for using the new task priority (old queues break because they weren't set up to support priority). Tangential idea, maybe we can add a flag `omit-queue-tag` or something if a user wants to run without this.

lucpeterson: alternatively, we could have this be a config setting (in the config file) with  default value that we change when we need to. then if we get a corner case where somebody really needs to go back and do something else, they can, without us having to explicitly check for it

lucpeterson: i kind of like this approach: have QUEUE_TAG be a configuration setting. most users will be fine with the default, which we can change as we see the need to do so, but also gives super-users the chance to revert or do something more wild for their own needs

ben-bay: Sounds good, I will implement this.

ben-bay: @lucpeterson I just added support for queue_tag in the config. The vpn from my personal machine is finicky so I haven't tested the distributed version of this, but let me know if this follows your feature idea.

ben-bay:
(Currently fixing a premature import)

merlin/celery.py
```diff
@@ -82,7 +82,7 @@
 
 # set task priority defaults to prioritize workflow tasks over task-expansion tasks
 task_priority_defaults = {
-    "task_queue_max_priority": get_priority(Priority.high),
+    "task_queue_max_priority": 10,
```
lucpeterson, line 85: are we sure about this?

ben-bay: I haven't tested this with a redis broker, but my gut feeling is that when asking for "max priority", it wants the highest integer value, to know how many queues to make.

ben-bay:
This PR​​​​​​​ looks good to go to me. I verified that it works on rabbit, with and without `omit_queue_tag` in the config file. Worked on first runs, second run w/ same workers and queues, and wf restarts. Also config customization produces the expected queue names.

PR #305, opened 2021-06-08
Merge the new CI that uses Github Actions

AlexanderWinterLLNL:
This would merge the new CI running on Github Actions into the develop branch of the main Merlin repo.

Modifications: the CI now runs on Github Actions. Certain tests and source code were modified to agree with the style standards that were indicated in documentation as required but not enforced.

TODO: while some modifications were made, bringing the entire code base up to standard would go well beyond the scope of the original ticket. I put some work into that refactor before deciding it would clutter this ticket too much of the core objective, the CI swap. While the CI flows over the existing linters, it abides by the standard that has existed up to this point, which is that merges will not fail out for being noncompliant with style standards. This will be changed in the next merge/ticket.

As discussed in the last meeting, bringing Merlin and the Spellbook up to standard will go through 3 phases: swapping the CI away from Travis to Actions; enforcing whatever standards were laid out originally and refactoring the codebase to reflect those standards, including adding PyLint and greater documentation, potentially type hinting and annotations; refactor the testing to be less idiosyncratic and operate consistently within the PyTest framework and automate deployment.


merlin/spec/specification.py
```diff
@@ -230,69 +230,8 @@ def dump(self):
...
-                        string += list_offset + (tab * lvl)
-                    else:
-                        string += tab * (lvl + 1)
-                    string += str(k) + ": " + str(dict_to_yaml(v, "", key_stack)) + "\n"
-                    key_stack.pop()
-                    i += 1
-            return string
-
-        result = dict_to_yaml(self.yaml_sections, "", [])
+        tab = 3 * " "
```
lucpeterson, line 233: tab is 3 spaces? I thought it's usually 4...

merlin/study/celeryadapter.py
```diff
@@ -229,7 +229,7 @@ def start_celery_workers(spec, steps, celery_args, just_return_command):
             if not check_machines(worker_machines):
                 continue
 
-            if yenv:
+            elif yenv:
```
lucpeterson, line 232: why this change? does it matter?

merlin/study/celeryadapter.py
```diff
@@ -352,6 +324,37 @@ def start_celery_workers(spec, steps, celery_args, just_return_command):
...
+            LOG.warning(
+                "The worker arg -O fair is recommended when running "
+                "parallel tasks"
+            )
+
+    if "-n" not in worker_args:
+        nhash = ""
+        if overlap:
+            nhash = time.strftime("%Y%m%d-%H%M%S")
+            # TODO: Once flux fixes their bug, change this back to %h
```
lucpeterson, line 351: unindent this comment, since it refers to the following line

tests/integration/run_tests.py
```diff
@@ -607,13 +607,13 @@ def define_tests():
...
                     negate=True,
                 ),
-                StepFileExistsCond(
-                    "verify",
-                    "MERLIN_FINISHED",
-                    "feature_demo",
-                    OUTPUT_DIR,
-                    params=True,
-                ),
+                # StepFileExistsCond(
```
lucpeterson, line 610: i thought we fixed this test?

merlin/spec/specification.py
```diff
@@ -301,6 +240,85 @@ def dict_to_yaml(obj, string, key_stack, newline=True):
...
+    def _process_string(self, obj, lvl, tab):
+        """
+        Processes strings for _dict_to_yaml() in the dump() method.
+        """
+        split = obj.splitlines()
+        if len(split) > 1:
+            obj = "|\n" + tab * (lvl + 1) + ("\n" + tab * (lvl + 1)).join(split)
+        return obj
+
+    def _process_dict_or_list(self, obj, string, key_stack, lvl, tab):
```
lucpeterson, line 270: i find this function hard to follow. i know it was hard to follow before and maybe i'm just staring at it for the first tie

Makefile
```diff
@@ -37,7 +37,8 @@ MRLN=merlin
 TEST=tests
 DOCS=docs
 WKFW=merlin/examples/workflows/
-MAX_COMPLEXITY?=5
+MAX_COMPLEXITY?=10
+MAX_LINE_LENGTH=127
```
ben-bay, line 41: fyi we've got another max line length in `setup.cfg`, so we should probably reconcile those.
https://github.com/LLNL/merlin/blob/9ea130dd83f67414a4759d331b944e5de3dc5ff7/setup.cfg#L14

merlin/spec/specification.py
ben-bay: Yeah, no idea what the convention is in yaml. Whatever is prettier

.github/workflows/push-pr_workflow.yml
```diff
@@ -0,0 +1,49 @@
+name: Python CI
```
lucpeterson, line 1: This looks like it has the right components, but I don't see that GitHub actions is actually being triggered on this PR. How can we verify that this is right? Do we need to enable actions in this repo?

merlin/study/celeryadapter.py
ben-bay: It looks like it won't change the logic since its conditional predecessor is a `continue`.

merlin/spec/specification.py
ben-bay: @AlexanderWinterLLNL I'm in favor of decomposition, but both versions are hard to follow at a glance. So can you verify that the "pretty" provenence spec generated by your changes is the same as before your changes?

merlin/spec/specification.py
ben-bay: @AlexanderWinterLLNL I'm in favor of decomposition, but both versions are hard to follow at a glance. So can you verify that the "pretty" provenence spec generated by your changes is the same as before your changes?

ben-bay: @AlexanderWinterLLNL I just ran this test on my personal machine w/ feature_demo, and both version indeed produce  identical provenance specs. Good job

.github/workflows/push-pr_workflow.yml
lucpeterson: We may have to disable the travis server to get this to work

ben-bay: Wouldn't this first need to be merged to `develop` or `main` to take over the CI?

merlin/study/celeryadapter.py
koning: The ```if yenv``` is actually separate logic from the continue on the check machines.

ben-bay: Uh oh I was wrong, it looks like we should indeed revert this.

Makefile
ben-bay: Have you thought of a way to do this?

.github/workflows/push-pr_workflow.yml
AlexanderWinterLLNL: Yes, I discussed this on the teams chat (guessing you didn't see it with your work machien in the shop) - it does need to be merged first. We'll do that and check the pipeline after merging, should be fine since [the fork ran the pipeline just fine and had no problems](https://github.com/AlexanderWinterLLNL/merlin/actions/runs/919970896)

Makefile
AlexanderWinterLLNL: Yeah, I split the variables assigned at the top of the Makefile into a config.mk, and have that file and the CI draw from setup.cfg as the source of all configurable settings.

merlin/study/celeryadapter.py
AlexanderWinterLLNL: You weren't wrong Ben, the logic is equivalent, this is just a trivial style preference I have. I'll revert it because that seems to be what the team prefers :)

lucpeterson:
So is this about good to go? What else do we need to include?

PR #306, opened 2021-06-14
Simple test reorganization

ben-bay:
Simple changes that make the `tests` module easier to understand from a developer perspective. No altered functionality. Should be a helpful starting point for future test refactoring.

* Added `unit` dir alongside existing `integration` dir and moved unit test files into it.
* Decomposed giant file `run_tests.py` into `conditions.py`, `define.py`, and `run_tests.py`.


tests/integration/conditions.py
```diff
@@ -0,0 +1,207 @@
...
+        :param `text`: text in which to search for a regex match
+        """
+        return search(self.regex, text) is not None
+
+    @property
+    def passes(self):
+        return self.contains()
+
+
+class ProvenanceCond(RegexCond):
```
lucpeterson, line 167: I think we can improve the name of this object in a future change (eg ProvenanceHasCond or ProvenanceContainsCond).

We could also consider dropping "Cond" from the names and making them more descriptive. 

Right now I understand this is just a re-org PR so don't do any of this here, but let's considering knocking some of this technical debt later.

tests/integration/define.py
```diff
@@ -0,0 +1,324 @@
+from conditions import (
```
lucpeterson, line 1: I think we should change the name of this file.
'define' is too generic and could clash with other libraries (maybe? down the road). maybe test_definitions.py

tests/integration/define.py
```diff
@@ -0,0 +1,324 @@
...
+    simple = f"{examples}/simple_chain/simple_chain.yaml"
+    slurm = f"{examples}/slurm/slurm_test.yaml"
+    slurm_restart = f"{examples}/slurm/slurm_par_restart.yaml"
+    flux = f"{examples}/flux/flux_test.yaml"
+    flux_restart = f"{examples}/flux/flux_par_restart.yaml"
+    lsf = f"{examples}/lsf/lsf_par.yaml"
+    black = "black --check --target-version py36"
+    config_dir = "./CLI_TEST_MERLIN_CONFIG"
+
+    return {
```
lucpeterson, line 41: I think this file could be significantly improved even more by breaking up the dictionary into thematic sub dictionaries and then grouping them together. For instance we could so something like:

```
merlin_checks = {"merlin": ..., "merlin help": ..., "merlin version": ..., "merlin config": ...}
local_workflows = {"local minimum_format:", ..., etc}
worker_command_checks = {"run-workers echo simple_chain:", ..., etc}
variable_override_checks = {...}

all_tests = {}
for d in [merlin_checks, local_workflows, worker_command_checks, variable_override_checks,]:
    _ = all_tests.update(d)

return all_tests
```

tests/integration/run_tests.py
```diff
@@ -33,19 +33,13 @@
...
 from contextlib import suppress
-from glob import glob
-from re import search
 from subprocess import PIPE, Popen
 
-from merlin.utils import get_flux_cmd
-
-
-OUTPUT_DIR = "cli_test_studies"
+from define import OUTPUT_DIR, define_tests
```
lucpeterson, line 42: see comment about renaming this

tests/integration/conditions.py
ben-bay: Okay, I can make an issue for this.

lucpeterson (approved):

Issue #275, opened 2020-08-29
[FEAT] unify walltime syntax


lucpeterson:
## 🚀 Feature Request

**What problem is this feature looking to solve?**
The different script adapters require different syntax for the wall time entry. If the time directive is in the wrong format, Merlin will fail to write the $(LAUNCHER) command, but with a mysterious error.

**Describe the solution you'd like**
We should have a unified interface and error catching if the entry is ill-formatted.

**Describe alternatives you've considered**
Leave it as is and let users figure it out.

koning:
There is only one time format that merlin accepts, the script adapters change it to the proper format.

PR #316, opened 2021-06-28
Feature/walltime syntax

lucpeterson:
Addresses #275 by allowing steps to be written in HH:MM:SS format, regardless of the batch type. Previously the script adapters had assumed formats. Also HH and MM are optional.


ben-bay (approved):
Looks good to me. We just need a CHANGELOG note.

ben-bay:
You're too quick!

lucpeterson:
> You're too quick!

#I'mTheBoss

koning (approved):

tests/unit/utils/test_time_formats.py
```diff
@@ -0,0 +1,68 @@
...
+        ("1:00", "00:01:00", "HMS"),
+        ("1:00", "60.0s", "FSD"),
+        ("1", "00:00:01", "HMS"),
+        ("1", "1.0s", "FSD"),
+    ]
+    for time_string, expected_result, method in cases:
+        check_convert_explicit(time_string, expected_result, method)
+
+
+def test_convert_timestring_same() -> None:
```
AlexanderWinterLLNL, line 42: I'd refactor this and the next test, and also the previous. Rather than using data internal to the unit tests, I'd leverage either Pytest's Fixtures or Pytest's mark.parametrize, and also merge this and the next unit test, especially since they're both going through check_all_equal() to examine results, and just include the rubric with the test data.

The benefit is A compartmentalization, data is separated from test structure, which more resembles the runtime environment, and B if something fails out, a fixture or mark.parametrize decorator still runs against all the parameters independently, whereas internal data like this will fail dependently - test case 3 could've passed or failed, but we won't know because case 2 failed and stopped the test prematurely for all parameters, we'd have to fix test 2 and run again from the top to get results for tests beyond 2 and just hope no other tests break.

The argument against is: this is done and works :D

Let me know which you'd prefer if you'd want to explore more  - fixtures are generally meant to be recycled across different tests (and especially different related or dependent modules), marks are fixed to the test they decorate. I generally prefer Fixtures, but marks are very useful when you to limit the scope and a bit more readable since they're attached exactly where they're needed.

tests/unit/utils/test_time_formats.py
```diff
@@ -0,0 +1,68 @@
...
+
+from merlin.utils import convert_timestring
+
+
+def _list_all_equal(my_list: List) -> bool:
+    """Boolean check that all items in the list are the same"""
+    return all(item == my_list[0] for item in my_list)
+
+
+def check_all_equal(
```
AlexanderWinterLLNL, line 11: should this and some of the other utility functions also be protected like _list_all_equal() for style? I guess I don't follow what distinguishes the protected and public methods.

AlexanderWinterLLNL (commented):
Feedback on the type hinting and Pytest environment! Looks mostly good, just a few suggestions.

ben-bay:
@AlexanderWinterLLNL do you see us adding type annotations to our codebase before the next release?

AlexanderWinterLLNL:
> @AlexanderWinterLLNL do you envision us adding type annotations to our codebase before the next release?

I think 1.8.0 could go out once this is merged, the codebase is large enough that adding type hints _everywhere_ would be spread across several tickets and pretty big, and almost certainly require MyPy to be activated in the CI to make sure they're consistent throughout. 

That said, if we're following semantic versioning, releasing a version with type hints would be more than a patch update, so if we released 1.8 we might move on to 1.9 sooner than usual.

merlin/utils.py
```diff
@@ -457,3 +459,75 @@ def contains_shell_ref(string):
...
+
+# Time utilities
+def convert_to_timedelta(timestr: str) -> timedelta:
+    """Convert a timestring to a timedelta object.
+    Timestring is given in in the format '[days]:[hours]:[minutes]:seconds'
+    with days, hours, minutes all optional add ons.
+    """
+    # make sure it's a string
+    timestr = str(timestr)
+    nfields = len(timestr.split(":"))
```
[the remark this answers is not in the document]
ymubarka: Maybe dumb question: How would you declare the type for `nfields`?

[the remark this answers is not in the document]
ymubarka: Never-mind I saw the next comment

tests/unit/utils/test_time_formats.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def check_all_equal(time_strings: List[Union[str, int]], method: Optional[str] = "HMS") -> bool:
+    """Check that time strings in a List convert to the same for the given conversion method."""
+    converted: List[str] = [convert_timestring(ts) for ts in time_strings]
+    all_equal: bool = _list_all_equal(converted)
+    return all_equal
+
+
+@pytest.mark.parametrize("time_string, expected_result, method", [
```
lucpeterson, line 20: this is super cool. a great improvement!

tests/unit/utils/test_time_formats.py
```diff
@@ -0,0 +1,50 @@
...
+    ("1:00", "00:01:00", "HMS"),
+    ("1:00", "60.0s", "FSD"),
+    ("1", "00:00:01", "HMS"),
+    ("1", "1.0s", "FSD")]
+)
+def test_convert_explicit(time_string: str, expected_result: str, method: str) -> None:
+    """Test some time strings to make sure they are converted correctly."""
+    converted: str = convert_timestring(time_string, method)
+    err_msg: str = f"Failed on time_string '{time_string}' for {method}, result was '{converted}', not '{expected_result}'"
+    assert converted == expected_result, err_msg
```
lucpeterson, line 32: nice use of the failure message. i didn't know you could do this

tests/unit/utils/test_time_formats.py
AlexanderWinterLLNL: Yeah, PyTest has some great builtin decorators and utilities!

tests/unit/utils/test_time_formats.py
AlexanderWinterLLNL: Yeah, really handy to be able to push whatever you want into stdout after a failed test!

Issue #319, opened 2021-06-30
[FEAT] [BUG] Change default queue_tag


lucpeterson:
## 🚀 Feature Request

**What problem is this feature looking to solve?**
The default queue tag [merlin]_ can cause problems with the run-workers command, since the command line has a [ in it. In some shells this causes the command to exit early and therefore the workers won't connect properly to the queue.

**Describe the solution you'd like**
Change the default queue tag to just merlin_

**Describe alternatives you've considered**
Could also (perhaps should) protect the queue name by putting the whole thing in quotes.

lucpeterson:
Resolved in #320

PR #320, opened 2021-06-30
Explictly wrap queue name string in quotes to make it more robust

lucpeterson:
Addresses #319 by wrapping the queue names in quotes.


ben-bay (approved):
We haven't released 1.8.0 yet so let's stick this fix under 1.8.0 on the CHANGELOG.

PR #324, opened 2021-07-16
Re-added support for all brokers with task priority

lucpeterson:
Fixes a bug introduced in 1.8.0 that inadvertently drops support for many of our supported brokers, included amqp, amqps.

This fix is needed to run on Summit / ORNL

### DIFF
--- a/merlin/config/utils.py
+++ b/merlin/config/utils.py
@@ -8,6 +8,11 @@ class Priority(enum.Enum):
     mid = 2
     low = 3
 
+def is_rabbit_broker(broker):
+    return broker in ["rabbitmq", "amqps", "amqp"]
+
+def is_redis_broker(broker):
+    return broker in ["redis", "rediss", "redis+socket"]
 
 def get_priority(priority):
     broker = CONFIG.broker.name.lower()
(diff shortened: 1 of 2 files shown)


koning (approved):

merlin/config/utils.py
```diff
@@ -8,6 +8,11 @@ class Priority(enum.Enum):
     mid = 2
     low = 3
 
+def is_rabbit_broker(broker):
```
AlexanderWinterLLNL, line 11: Was PyLint complaining about complexity or line length for having these 1 line `in` examinations in the body of get_priority()? Unless we anticipate maybe growing the supported brokers (do we?), this seems a simple enough check we could embed it in get_priority() without a function lookup, but it's just stylistic difference, looks good otherwise.

AlexanderWinterLLNL (approved):
Had one question/observation, but otherwise LGTM.

merlin/config/utils.py
lucpeterson: this is mostly to future-proof against possible other brokers in the future. we have no plans, but celery supports a bunch of different ones.

lucpeterson: also a bunch of different kinds of brokers can be classified as either rabbit or redis brokers (eg depending on ssl), so it's nice to be able to add them here

PR #326, opened 2021-08-05
Refactor of Merlin internals and CI to be flake8 compliant

AlexanderWinterLLNL:
This could've been broken into several PRs, although most of the files touched were cosmetic (removing stray spaces, moving away from 2.X conventions for 3.3> conventions, etc).

Notable changes: 

- a number of files were updated to be flake8 compliant, several larger functions needed to be reduced in complexity.
  - one function, merlin_step in tasks.py was irreducible and had to be overstepped - the celery decorator and global behavior from it was causing problems when dispatched into smaller functions.
  - anything that was touched was typically refactored to have type annotations.
- a number of Makefile targets weren't configured to properly use the venv, every statement of the target moves through a subshell so the venv would effectively be dropped and dependencies installed outside the venv to the system instead. Makefile targets were updated and now move through a single subshell.
- continuous integration was updated to reflect the style standards so all future PRs are flake8 compliant.


merlin/study/batch.py
```diff
@@ -110,16 +111,19 @@ def get_node_count(default=1):
...
     """
     The configuration in the batch section of the merlin spec
     is used to create the worker launch line, which may be
     different from a simulation launch.
 
-    com (str): The command to launch with batch configuration
-    nodes (int): The number of nodes to use in the batch launch
-    batch (dict): An optional batch override from the worker config
-
+    : param spec : I don't know
```
lucpeterson, line 123: workflow specification

merlin/study/celeryadapter.py
```diff
@@ -260,7 +245,11 @@ def start_celery_workers(spec, steps, celery_args, just_return_command):
         # Add a per worker log file (debug)
         if LOG.isEnabledFor(logging.DEBUG):
             LOG.debug("Redirecting worker output to individual log files")
-            worker_args += f" --logfile %p.%i"
+            # The linter pointed out this f string currently does nothing, but
+            # this looks like this was swapped between % formatting and f-strings.
+            # Current implementation will just print '%p.%i', if that is not intended
+            # it needs to be updated. Currently skipping linter examination.
```
lucpeterson, line 251: the f can be removed. the intent is to print %p.%i . these get passed to the celery workers and are interpreted there

lucpeterson:
Goodness that's quite the PR! Well done. Get another pair of eyes on it. The part I'm least sure about is the Makefile changes, but the rest looks like it'll work

merlin/config/utils.py
```diff
@@ -8,17 +9,20 @@ class Priority(enum.Enum):
...
+def is_redis_broker(broker: str) -> bool:
     return broker in ["redis", "rediss", "redis+socket"]
 
-def get_priority(priority):
-    broker = CONFIG.broker.name.lower()
-    priorities = [Priority.high, Priority.mid, Priority.low]
-    if priority not in priorities:
-        raise ValueError(
+
+def get_priority(priority: Priority) -> int:
```
lucpeterson, line 21: this is called eg via get_priority(Priority.high) or get_priority(Priority.low). Are these Priority objects? Enums are weird sometimes.

merlin/examples/workflows/flux/scripts/flux_info.py
```diff
@@ -68,24 +69,28 @@
 except BaseException:
     top_dir = "job"
 
-    def get_data_dict(key):
-        kwargs = {
+    def get_data_dict(key) -> Dict:
```
lucpeterson, line 72: i think key is a str

merlin/examples/workflows/flux/scripts/flux_info.py
```diff
@@ -81,10 +81,10 @@ def get_data_dict(key):
         stdout, stderr = p.communicate()
 
         data = {}
-        for l in stdout.split("/n"):
-            for s in l.strip().split():
-                if "timestamp" in s:
-                    jstring = s.replace("'", '"')
+        for line in stdout.split("/n"):
```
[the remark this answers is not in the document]
lucpeterson: it's ok to skip completing this for now. it's low priority (example) code that isn't really used in production. if we get distributed tests up we might use it more, but I don't even know if it works or not

merlin/study/batch.py
```diff
@@ -143,60 +147,70 @@ def batch_worker_launch(spec, com, nodes=None, batch=None):
...
+            f"{launches} {flux_exe} start {flux_opts} {flux_exec} `which {shell}` -c"
         )
         worker_cmd = f'{launch} "{com}"'
+    else:
+        worker_cmd = f"{launches} {com}"
 
     return worker_cmd
+
+
+def construct_launches(batch: Dict, btype: Union[str, Dict], nodes: int) -> str:
```
lucpeterson, line 191: maybe we can rename this construct_worker_launch_command

merlin/study/batch.py
```diff
@@ -110,16 +111,19 @@ def get_node_count(default=1):
     return default
 
 
-def batch_worker_launch(spec, com, nodes=None, batch=None):
+def batch_worker_launch(spec: Dict,
+                        com: str,
+                        nodes: Union[str, Dict] = None,
+                        batch=None) -> str:
```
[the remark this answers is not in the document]
AlexanderWinterLLNL: Hmm - the docstring's untouched from main and as is only indicates the desired type, not accepted types the variable is expected to possibly hold when passed or during the function, which is what's needed (if datatype changes between passing and what's handled internally, I prefer two variables - one for better type enforcement, to make sure what's passed is valid, and one that's defined in while processed the passed var to whatever type is needed elsewhere).

Seeing it again, both the docstring and signature are wrong, this should be Optional because None is valid, the union is missing int, and Dict isn't valid. Mypy would catch those discrepancies when we bring it online.

I'll fix both.

merlin/config/utils.py
AlexanderWinterLLNL: Yeah, they can be funky. If passed that way it'll be a Priority, although if it were Priority.high.name or Priority.high.val it'd be a str or int. Could more broadly type it as an Enum, but that would actually fail on something like `assert type(priority_obj) == Enum)` (you'd have to use isinstance() to get the parent Enum for Priority). Generally safer to be as specific as possible unless you have a specific reason to be generic (like a factory method needing to handle different classes).

merlin/common/opennpylib.py
```diff
@@ -280,9 +281,10 @@ def to_array(self):
 
 
 class OpenNPYList:
-    def __init__(self, l):
-        self.filenames = l
-        self.files = [OpenNPY(_) for _ in self.filenames]
+    def __init__(self, filename_strs: List[str]):
+        self.filenames = filename_strs
```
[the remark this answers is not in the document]
AlexanderWinterLLNL: Updated.

merlin/common/opennpylib.py
```diff
@@ -280,9 +281,10 @@ def to_array(self):
...
+        self.files: List[OpenNPY] = [OpenNPY(_) for _ in self.filenames]
+        i: OpenNPY
         for i in self.files:
             i.load_header()
         self.shapes = [_.hdr["shape"] for _ in self.files]
```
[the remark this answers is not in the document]
AlexanderWinterLLNL: Updated.

Makefile
```diff
@@ -29,112 +29,86 @@
...
-# this only works outside the venv
+# this only works outside the venv - if run from inside a custom venv, or any target that depends on this,
+# you will break your venv.
 virtualenv:
-	$(PYTHON) -m venv $(VENV) --prompt $(PENV) --system-site-packages
-	$(PIP) install --upgrade pip
-
+	(
+	    $(PYTHON) -m venv $(VENV) --prompt $(PENV) --system-site-packages; \
+	    . $(VENV)/bin/activate; \
```
lucpeterson, line 57: this might not work in tcsh

merlin/examples/workflows/flux/scripts/flux_info.py
lucpeterson: change to key:str

merlin/study/batch.py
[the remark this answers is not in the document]
lucpeterson: the nodes field in the header should be an int?

merlin/study/batch.py
```diff
@@ -143,60 +147,70 @@ def batch_worker_launch(spec, com, nodes=None, batch=None):
...
-                launchs += f" -p {queue}"
-            if walltime:
-                launchs += f" -t {walltime}"
-        if launcher == "lsf":
-            # The jsrun utility does not have a time argument
-            launchs = f"jsrun -a 1 -c ALL_CPUS -g ALL_GPUS --bind=none -n {nodes}"
-
-    launchs += f" {launch_args}"
+    if not launches:
+        launches = construct_launches(batch, btype, nodes)
```
lucpeterson: Oh I see now. 'launches' isn't a good variable name. it's supposed to be a stand in for 'launch_string" (it was 'launchs' before, which wasn't a misspelling of 'launches').
Please change "launches" to launch_string or launch_command and change the "construct_launches" function to be "construct_launch_string" or "construct_launch_command"

merlin/study/batch.py
```diff
@@ -143,60 +147,70 @@ def batch_worker_launch(spec, com, nodes=None, batch=None):
...
-    shell = get_yaml_var(batch, "shell", "bash")
-    walltime = get_yaml_var(batch, "walltime", "")
+    shell: Union[str, Dict] = get_yaml_var(batch, "shell", "bash")
 
-    launch_pre = get_yaml_var(batch, "launch_pre", "")
-    launch_args = get_yaml_var(batch, "launch_args", "")
-    worker_launch = get_yaml_var(batch, "worker_launch", "")
+    launch_pre: Union[str, Dict] = get_yaml_var(batch, "launch_pre", "")
+    launch_args: Union[str, Dict] = get_yaml_var(batch, "launch_args", "")
+    launches: str = get_yaml_var(batch, "worker_launch", "")
```
lucpeterson, line 156: change name per below

merlin/study/batch.py
```diff
@@ -143,60 +147,70 @@ def batch_worker_launch(spec, com, nodes=None, batch=None):
...
-                launchs += f" -t {walltime}"
-        if launcher == "lsf":
-            # The jsrun utility does not have a time argument
-            launchs = f"jsrun -a 1 -c ALL_CPUS -g ALL_GPUS --bind=none -n {nodes}"
-
-    launchs += f" {launch_args}"
+    if not launches:
+        launches = construct_launches(batch, btype, nodes)
+
+    launches += f" {launch_args}"
```
lucpeterson, line 161: change variable name throughout this function

merlin/study/batch.py
lucpeterson: see above comments. change the name and make it consistent

merlin/study/batch.py
```diff
@@ -143,60 +147,70 @@ def batch_worker_launch(spec, com, nodes=None, batch=None):
...
+        launches = f"srun -N {nodes} -n {nodes}"
+        if bank:
+            launches += f" -A {bank}"
+        if queue:
+            launches += f" -p {queue}"
+        if walltime:
+            launches += f" -t {walltime}"
+    if launcher == "lsf":
+        # The jsrun utility does not have a time argument
+        launches = f"jsrun -a 1 -c ALL_CPUS -g ALL_GPUS --bind=none -n {nodes}"
```
lucpeterson, line 216: this should be 'launch_string"

merlin/study/batch.py
```diff
@@ -143,60 +147,70 @@ def batch_worker_launch(spec, com, nodes=None, batch=None):
...
+def construct_launches(batch: Dict, btype: Union[str, Dict], nodes: int) -> str:
+    """
+    If no 'worker_launch' is found in the batch yaml, this method constructs the needed data.
+
+    : param batch : (dict): An optional batch override from the worker config
+    : param btype : (str): The type of batch (flux, local, lsf)
+    : param nodes : (int): The number of nodes to use in the batch launch
+    """
+    launches: str = ""
+    launcher: str = get_batch_type()
```
lucpeterson, line 202: 'launch string"

merlin/study/batch.py
```diff
@@ -143,60 +147,70 @@ def batch_worker_launch(spec, com, nodes=None, batch=None):
...
+
+def construct_launches(batch: Dict, btype: Union[str, Dict], nodes: int) -> str:
+    """
+    If no 'worker_launch' is found in the batch yaml, this method constructs the needed data.
+
+    : param batch : (dict): An optional batch override from the worker config
+    : param btype : (str): The type of batch (flux, local, lsf)
+    : param nodes : (int): The number of nodes to use in the batch launch
+    """
+    launches: str = ""
```
lucpeterson, line 201: why the blank str and then override it one line later?

Makefile
lucpeterson: i actually don't think you need to 'activate' the venv, b/c the $(PIP) string points to the pip inside that venv already

Makefile
```diff
@@ -29,115 +29,90 @@
...
+	)
 
 
-# remove doc build files
-clean-docs:
-	rm -rf $(DOCS)/build
+install-workflow-deps: virtualenv install-merlin
+	( \
+	   . $(VENV)/bin/activate; \
+	   merlin example feature_demo; \
```
lucpeterson, line 75: i don't think you need to do this, b/c won't $(WKFW) point directly to where the example is in the source file?

merlin/study/batch.py
AlexanderWinterLLNL: ahhh! Ok, I'll update it.

AlexanderWinterLLNL: And construct_launches() is now construct_worker_launch_command().

merlin/study/batch.py
AlexanderWinterLLNL: The names are similar but they are different strings (although as this is pointing out, probably need to be more different). The names are the same as main (other than the mistake around launchs) - `launches` should be `launch_string` as you pointed out, but `launcher` is about the distributed computing manager. It's used in some logic below to check whether to run slurm, or flux, or lsf, and builds `launch_string` differently depending on which.

Maybe `launcher` ought to be `workload_manager`?

merlin/study/batch.py
AlexanderWinterLLNL: I don't think so? See the above comment, this is logic sorting which distributed workload manager the `launch_string` needs and builds the string differently depending on which is selected, `launcher` holds which DWM is needed.

merlin/study/batch.py
AlexanderWinterLLNL: Patched.

merlin/study/batch.py
AlexanderWinterLLNL: Patched.

Makefile
AlexanderWinterLLNL: Oh, yes! I just noticed the disparity between the Make targets and the CI and was just synching them up - didn't check the config vars that closely when I pulled them out :D

Makefile
AlexanderWinterLLNL: oh, I didn't notice that! I'll remove them - this fell out of automating testing, bc those targets fail without activating it. I don't remember what prompted the thought, but the concern was that running the install targets without activating (pathing pip to the venv) was just installing to the environment if we didn't guaranteeing the venv, and would then fail. Should've looked at the config vars more closely.

merlin/study/batch.py
AlexanderWinterLLNL: I don't think so? See [this comment](https://github.com/LLNL/merlin/pull/326#discussion_r701470193), this is logic sorting which distributed workload manager the `launch_string` needs and builds the string differently depending on which is selected, `launcher` holds which DWM is needed.

merlin/study/batch.py
lucpeterson: gah! i see i misread the character difference

Makefile
lucpeterson: yeah i think it's the config variables that you can use to change this behavior if needed

lucpeterson:
ok let me know when you push those patches and if you need me to look again in the morning. i think we're on track to include this with 1.8.1

merlin/study/batch.py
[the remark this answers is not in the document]
AlexanderWinterLLNL: Patched.

merlin/study/batch.py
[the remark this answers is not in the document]
AlexanderWinterLLNL: It's needed as an int but at lines 142 and 147 `nodes` can be a None, int, or str. We can declare it only an int in the header/signature, but if/when we do type enforcement with mypy it would fail out at those lines.

I added a check to make sure that, at the point `nodes` is supposed to have been processed to an int, it is actually an int.

merlin/examples/workflows/flux/scripts/flux_info.py
AlexanderWinterLLNL: I changed this the day the comment was made, it shows it correctly in the Files changed tab, but it's not updating here, the comments are remaining tied to the commit they were made with and not updating to the code they're commenting on in the current commit.

lucpeterson (approved):

PR #331, opened 2021-09-13
Pylint ci refactor

AlexanderWinterLLNL:
This has the first handful of PyLint refactored files. This refactor is too large for one pass, so it will be broken up across several PRs. Like the Flake8 refactor, any classes/functions that were touched by this were (generally) given type hints to make the transition to enforcing MyPy easier when that comes online.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,12 +32,15 @@ include config.mk
 .PHONY : virtualenv
 .PHONY : install-merlin
 .PHONY : install-workflow-deps
-.PHONY : install-merlin-dev
+.PHONY : install-dev
 .PHONY : unit-tests
 .PHONY : e2e-tests
 .PHONY : tests
-.PHONY : fix-style
+.PHONY : check-flake8
+.PHONY : check-black
+.PHONY : check-pylint
 .PHONY : check-style
+.PHONY : fix-style
 .PHONY : check-push
 .PHONY : check-camel-case
 .PHONY : checks
@@ -49,6 +52,7 @@ include config.mk
 .PHONY : clean-py
 .PHONY : clean
 
+
 # this only works outside the venv - if run from inside a custom venv, or any target that depends on this,
 # you will break your venv.
 virtualenv:
@@ -56,8 +60,10 @@ virtualenv:
 	$(PIP) install --upgrade pip; \
 	$(PIP) install -r requirements/release.txt; \
 
+
 # install merlin into the virtual environment
 install-merlin: virtualenv
+	. $(VENV)/bin/activate; \
 	$(PIP) install -e .; \
 	merlin config; \
 
@@ -68,7 +74,7 @@ install-workflow-deps: virtualenv install-merlin
 
 
 # install requirements
-install-merlin-dev: virtualenv install-workflow-deps
+install-dev: virtualenv install-workflow-deps
 	$(PIP) install -r requirements/dev.txt; \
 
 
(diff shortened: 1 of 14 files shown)


merlin/celery.py
```diff
@@ -131,10 +137,13 @@ def setup(**kwargs):
...
-        current = billiard.current_process()
-        prefork_id = current._identity[0] - 1  # range 0:nworkers-1
-        cpu_slot = (prefork_id * cpu_skip) % npu
-        p.cpu_affinity(list(range(cpu_slot, cpu_slot + cpu_skip)))
+        cpu_skip: int = int(os.environ["CELERY_AFFINITY"])
+        npu: int = psutil.cpu_count()
+        process: psutil.Process = psutil.Process()
+        # pylint is upset that typing accesses a protected class, ignoring W0212
+        # pylint is upset that billiard doesn't have a current_process() method - it does
+        # pylint: disable=W0212, E1101
```
KaseyNagleLLNL, line 145: Do we want to re -enable these after? Same for the disable on 130 (W0613)

merlin/main.py
```diff
@@ -68,33 +71,35 @@ def error(self, message):
         sys.exit(2)
 
 
-def verify_filepath(filepath):
+def verify_filepath(filepath) -> str:
```
KaseyNagleLLNL, line 74: filepath typing

merlin/main.py
```diff
@@ -68,33 +71,35 @@ def error(self, message):
...
     """
     Verify that the filepath argument is a valid
     file.
 
     :param `filepath`: the path of a file
```
KaseyNagleLLNL, line 79: Do we need to add return typing to documentation?  I saw a couple other places they werent added so I wasnt sure.

merlin/main.py
```diff
@@ -68,33 +71,35 @@ def error(self, message):
...
     """
-    filepath = os.path.abspath(os.path.expandvars(os.path.expanduser(filepath)))
+    filepath: str = os.path.abspath(os.path.expandvars(os.path.expanduser(filepath)))
     if not os.path.isfile(filepath):
         raise ValueError(f"'{filepath}' is not a valid filepath")
     return filepath
 
 
-def verify_dirpath(dirpath):
+def verify_dirpath(dirpath) -> str:
```
KaseyNagleLLNL, line 87: dirpath typing

merlin/main.py
```diff
@@ -126,16 +132,16 @@ def parse_override_vars(variables_list):
...
 
-            val = entry[1]
+            val: Union[str, int] = entry[1]
             with suppress(ValueError):
                 int(val)
                 val = int(val)
             result[key] = val
 
-        except BaseException as e:
+        except BaseException as excpt:
```
KaseyNagleLLNL, line 141: Does excpt need typing?

merlin/main.py
```diff
@@ -297,25 +303,27 @@ def print_info(args):
...
-        USER_HOME = os.path.expanduser("~")
-        output_dir = os.path.join(USER_HOME, ".merlin")
-    _ = router.create_config(args.task_server, output_dir, args.broker)
+        user_home: str = os.path.expanduser("~")
+        output_dir: str = os.path.join(user_home, ".merlin")
+    router.create_config(args.task_server, output_dir, args.broker)
 
 
-def process_example(args):
+def process_example(args) -> None:
```
KaseyNagleLLNL, line 325: Args typing

merlin/main.py
```diff
@@ -735,10 +754,16 @@ def main():
 
     try:
         args.func(args)
-    except Exception as e:
+        # pylint complains that this exception is too broad - being at the literal top of the program stack,
+        # it's ok.
+    except Exception as excpt:  # pylint: disable=broad-except
```
KaseyNagleLLNL, line 759: Same excpt typing comment as before

merlin/celery.py
AlexanderWinterLLNL: Good catch, patched to disable-next.

merlin/main.py
AlexanderWinterLLNL: We should, I'm sure there are places I missed it. I'll make sure everything that was pushed with a ret val is declared.

merlin/main.py
AlexanderWinterLLNL: Patched.

merlin/main.py
AlexanderWinterLLNL: I don't believe so..? It's implicitly a BaseException, so, any Exception at all :D. I suppose it could be forward declared like 
```
excpt: BaseException
except BaseException as excpt:
```
but, I'm not sure if that would work with the context management of the `except` clause, bc I think that the forward declared var wouldn't actually be used. Kind of like forward declaring the temp var used in a list comprehension causes  "unused var" linter problems for the forward declaration.

merlin/main.py
AlexanderWinterLLNL: Patched.

merlin/main.py
AlexanderWinterLLNL: Patched.

lucpeterson:
This is looking pretty good

merlin/celery.py
```diff
@@ -97,15 +100,15 @@
...
 ):
     LOG.debug("Skipping celery config override; 'celery.override' field is empty.")
 else:
-    override_dict = nested_namespace_to_dicts(CONFIG.celery.override)
-    override_str = ""
-    i = 0
+    override_dict: Dict = nested_namespace_to_dicts(CONFIG.celery.override)
+    override_str: str = ""
+    i: int = 0
     for k, v in override_dict.items():
```
AlexanderWinterLLNL, line 112: Ideally these would have some sort of forward declaration, but I have no idea what they would be.

Makefile
```diff
@@ -85,23 +91,42 @@ e2e-tests:
...
+
+check-black:
+	. $(VENV)/bin/activate; \
+	$(PYTHON) -m black --check --target-version py36 $(MRLN); \
+
+
+check-pylint:
+	. $(VENV)/bin/activate; \
+	echo "PyLinting merlin source..."; \
+	$(PYTHON) -m pylint merlin --rcfile=setup.cfg --ignore-patterns="$(VENV)/" --disable=logging-fstring-interpolation; \
```
AlexanderWinterLLNL, line 118: As a note on what this is doing, it is: ignoring PyLint examination of the venv (we don't care if our dependencies are linted); ignoring f-string interpolation.

Pylint warns about using f-strings in Loggers because Loggers have a particular optimization about interpolating % strings (namely, it does not evaluate the string expansion fr a % str if a logging event is not encountered) which it does not have for f-strings (the f-string is evaluated whether or not the event is logged, which is pointless work and for certain log objects could have significant runtime).

We could add other global exceptions if anyone thinks of any, but given how much more readable f strings are, I thought this was an ok trade.

merlin/config/configfile.py
```diff
@@ -111,21 +111,21 @@ def load_default_user_names(config):
...
 
 
-def get_config(path):
+def get_config(path: Optional[str]) -> Dict:
     """
     Load a merlin configuration file and return a dictionary of the
     configurations.
 
-    :param path : The path to search for the config file.
+    :param [Optional[str]] path : The path to search for the config file.
```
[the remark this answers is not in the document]
AlexanderWinterLLNL: Patched.

KaseyNagleLLNL (approved):

PR #333, opened 2021-09-15
Enabled distributed testing suite.

AlexanderWinterLLNL:
Two changes. To the existing distributed test, added a `sleep` command and `stop-workers` so that the process could reach completion (prior to, the workers were simply left standing, awaiting additional tasks, so the test never 'completed' and just hung forever). The other, more important addition, was of another test, remote_feature_demo, which is simply the feature_demo with an additional step in the study, 'stop-workers', which 'exits' at the end of the workflow for a more graceful exit than the previous change.


tests/integration/test_definitions.py
```diff
@@ -327,8 +328,27 @@ def define_tests():
...
+                    "verify",
+                    "MERLIN_FINISHED",
+                    "feature_demo",
+                    OUTPUT_DIR,
+                    params=True,
+                ),
+            ],
+        ),
+        "stop workers for distributed feature_demo": (
+            f"{run} {demo} --vars OUTPUT_PATH=./{OUTPUT_DIR} WORKER_NAME=cli_test_demo_workers ; {workers} {demo} --vars OUTPUT_PATH=./{OUTPUT_DIR} WORKER_NAME=cli_test_demo_workers ; sleep 20 ; merlin stop-workers",
```
lucpeterson, line 351: i don't like having this test in here, because it will stop all of the workers the user has connected
i'd be ok with doing it if we did stop-workers --spec {demo}, but that's broken right now

tests/integration/test_definitions.py
```diff
@@ -362,7 +382,7 @@ def define_tests():
         # provenence_equality_checks, # omitting provenance equality check because it is broken
         # style_checks, # omitting style checks due to different results on different machines
         dependency_checks,
-        # distributed_tests, # omitting distributed tests as they are not yet ready
+        distributed_tests
```
lucpeterson, line 385: does the ci run this or just the --local tests? b/c we for distributed we don't have a ci version working (until we get a rabbit and/or redis docker pop-up server in the pipeline)

lucpeterson (changes requested):

tests/integration/test_definitions.py
AlexanderWinterLLNL: Pulled.

tests/integration/test_definitions.py
AlexanderWinterLLNL: It now runs the distributed tests and also then fails out (because this doesn't setup the docker containers). I'm looking into how to set that up with Actions.

tests/integration/test_definitions.py
```diff
@@ -316,13 +316,13 @@ def define_tests():
...
         "black check tests": (f"{black} tests/", HasReturnCode(), "local"),
     }
-    dependency_checks = {
-        "deplic no GNU": (
-            "deplic ./",
-            [HasRegex("GNU", negate=True), HasRegex("GPL", negate=True)],
-            "local",
-        ),
-    }
+    # dependency_checks = {
```
lucpeterson, line 319: why commented out?

merlin/examples/workflows/remote_feature_demo/scripts/pgen.py
```diff
@@ -0,0 +1,14 @@
+from maestrowf.datastructures.core import ParameterGenerator
+
+
+def get_custom_generator(env, **kwargs):
+    p_gen = ParameterGenerator()
```
KaseyNagleLLNL, line 5: Is typing needed here?

tests/integration/test_definitions.py
AlexanderWinterLLNL: Just a relic of experimenting with the pipeline, it's back now - the Actions log _seemed_ to indicate that some of the pipelines were hanging on the dependency check before they even got to the distributed test (it wasn't, it was hanging on the distributed test).

tests/integration/test_definitions.py
AlexanderWinterLLNL: The CI now stands up a docker image with a rabbitmq server, but it is not linked for the workflow to access, I'm still figuring out how to do that, but haven't spent any time on this since the last commit as other work's had priority. 

As a thought, this could be merged, with the CI only run against the local tests, and we'd just require any developer to run the full, distributed test suite against whatever cluster environment they developed on to verify nothing's been broken. That is not a long term solution, but 
A. it would only be a stopgap until I have time to come back and figure out the linking 
B. we've had a similar manual process as the working solution for black and isort until now (that said, I put up [this PR](https://github.com/LLNL/merlin/pull/336) to end that) 
C. most importantly, as-is this runs on LC machines, merging would enable testing against a distributed configuration, which is a critical and significant improvement, even if it isn't (yet) automated for the CI in our Actions environment.

.github/workflows/push-pr_workflow.yml
```diff
@@ -90,6 +91,78 @@ jobs:
...
+      run: |
+        python3 tests/integration/run_tests.py --verbose --ids 31 32
+
+    - name: Setup rabbitmq config
+      run: |
+        merlin config --broker rabbitmq
+
+    - name: Run integration test suite for rabbitmq
+      env:
+        RABBITMQ_HOST: redis
```
lucpeterson, line 165: is this intentional right now?

AlexanderWinterLLNL: No haha

AlexanderWinterLLNL:
So I think I figured out the problem with rabbitmq. When the config is initialized, the default app.yaml is hard-coded to expect Jackalope credentials, so, while there is a valid rabbitmq stood up in the CI environment which Merlin _could_ connect to, it never links because the runner hangs expecting jackalope credentials it doesn't have. Two possible solutions:

1. we provide the credentials with a [github secret](https://docs.github.com/en/actions/security-guides/encrypted-secrets). I don't advocate this, mostly because I'm not certain of the security and don't want to deploy anything which I'm not certain of the security - the only users who can access secrets are Maintainers, but, I'm not certain someone couldn't do something clever to, I don't know, ssh to Jackalope. I don't know how they could, but, I don't want to do something I'm not positive about when I know we could...
2. use the currently unused `merlin config --test` infrastructure. Currently `merlin config --test` is built in as a potential arg to `merlin config`, but, it's not actually invoked anywhere in the codebase (AFAIS), and, even when it is, all it does is update the log from 'Creating config...' to 'Creating test config...'. All we'd need to do is add another data config file ('test_app.yaml') in merlin.data.config for rabbitmq that gathers the necessary data from the os.environ, and the test should flow.

lucpeterson:
is the plan to push this PR through with just the redis ci and add rabbitmq in a separate PR?

PR #334, opened 2021-09-22
CI changes: separate linting and testing, add caching, add check for CHANGELOG update

AlexanderWinterLLNL:
This refactors the CI in three ways. This was based off [this Actions doc](https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows) and [this blog post](https://medium.com/ai2-blog/python-caching-in-github-actions-e9452698e98d)

1. The first separates linting and testing. A it's a more efficient use of the Actions runner to use more of the parallel resources we're allocated, significantly reducing CI runtime (about 25%) B. we shouldn't stop the whole CI pipeline because Flake8 noticed an extra end space, its better to know more potential failure points at each push (were there also problems in testing, PyLint, eventually MyPy, etc.)
2. Dependencies are loaded from a cache, greatly reducing the installation time across runs (both for the specific run and every succeeding run). The cache is keyed to the runtime environment and examined for updates to the dependent libraries each time it is stood up. This already has significant runtime improvement (about 25% again beyond the previous improvement) but this will be even more valuable when, e.g. Spellbook\* begins incorporating large dependencies like PyTorch and BoTorch.
3. We examine that every PR has at least some indication of the change it's making in the CHANGELOG.

There are some other changes we ought to make in the future - while Merlin supports a Docker image, currently we aren't examining that in our CI, everything is run against a venv. While not critical, it is possible backend updates to Docker, some minor change in our source, _something_ could break that, we should be running the project in that environment it as part of CI if we're supporting it.

\* I noticed Spellbook is not considered a dependency for dev - do we want to continue with this? It absolutely is a dev requirement because of the test suite's dependence on the example workflows. It has its own internal requirements that it installs, but, the fusion of the example workflows with the integrated workflows also makes it an explicit dev dependency.

### DIFF
--- a/.github/workflows/push-pr_workflow.yml
+++ b/.github/workflows/push-pr_workflow.yml
@@ -3,9 +3,43 @@ name: Python CI
 on: [push, pull_request]
 
 jobs:
+  lint:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.9]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install dependencies
+      run: |
+        python3 -m pip install --upgrade pip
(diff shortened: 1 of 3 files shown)


lucpeterson:
nice.

KaseyNagleLLNL (approved):

PR #335, opened 2021-09-24
'Shell console output fix

AlexanderWinterLLNL:
Patch to suppress console output about unsupported shell, minor fix to Makefile target.

Updated self._unsupported and new_unsupported in script_adapter.py to include 'shell', overcoming console arning about shell being unsupported. Typing added to touched class __inits__.
Makefile had an error in install-merlin that wasn't activating the venv, now activates to enable that target as well as install-dev.

### DIFF
--- a/merlin/study/script_adapter.py
+++ b/merlin/study/script_adapter.py
@@ -34,6 +34,7 @@ Merlin script adapter module
 
 import logging
 import os
+from typing import Dict, List, Set
 
 from maestrowf.interfaces.script import SubmissionRecord
 from maestrowf.interfaces.script.localscriptadapter import LocalScriptAdapter
@@ -66,7 +67,7 @@ class MerlinLSFScriptAdapter(SlurmScriptAdapter):
         """
         super(MerlinLSFScriptAdapter, self).__init__(**kwargs)
 
-        self._cmd_flags = {
+        self._cmd_flags: Dict[str, str] = {
             "cmd": "jsrun",
             "ntasks": "--np",
             "nodes": "--nrs",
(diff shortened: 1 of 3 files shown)


merlin/study/script_adapter.py
```diff
@@ -79,22 +80,23 @@ def __init__(self, **kwargs):
             "lsf": "",
         }
 
-        self._unsupported = {
+        self._unsupported: Set[str] = {
```
lucpeterson, line 83: now that i look at this. why is this a set and the other one (new_unsupported) a list?

lucpeterson (approved):

merlin/study/script_adapter.py
AlexanderWinterLLNL: Mm, I don't know ¯\\_(ツ)_/¯. AFAIS `new_unsupported` is only used in the constructor to build a new `_unsupported` set by union'ing its list of strs and then it's dumped. Maybe `_unsupported` needs to be guaranteed collision free somewhere else and that's why it's a Set (it's a Set because that's what it is in the parent class from Maestro), and `new_unsupported` is a List because it doesn't matter after construction?

I tend to make these sorts of symbols Enums rather than handle Lists/Sets of raw strings, although being fed these data types by inheritance from Maestro kind of ties hands.

AlexanderWinterLLNL: My confusion was why MerlinLSFScriptAdapter was dumping the inherited attributes it initializes in the parent and overwriting with entirely new, largely parallel data rather than updating the homologous data structs the parent already initialized like MerlinSlurmScriptAdapter does. It's nice MerlinLSFScriptAdapter makes it explicit what keys you could grab in Merlin from `_unsupported`, especially given its parent is a library that has to be accessed through linking/the venv, but, we might want to pick one or the other.

koning: The set was used to eliminate duplicates from the lists. The various unsupported lists were created to work around maestro issues. This can be refactored to create a generic parallel launcher with external configuration. This was one of the feature proposals near the end of the project.

PR #348, opened 2021-10-29
Fixing the 'merlin example list' error 

ymubarka:


lucpeterson (changes requested):
1. Please update the changeling
2. Line 66 in template_config.py will break because it will over-write the optimization.yaml file that you want to keep. Maybe rename the saving file something else?
3. Could you add a test to the integration tests for "merlin example list" to succeed without error, so that we catch any of these issues in the future?

ymubarka:
> 1. Please update the changeling
> 2. Line 66 in template_config.py will break because it will over-write the optimization.yaml file that you want to keep. Maybe rename the saving file something else?
> 3. Could you add a test to the integration tests for "merlin example list" to succeed without error, so that we catch any of these issues in the future?

When running the tests on my local machine it was giving me errors and stalling when it got to the second to last test. Is that normal?

lucpeterson (approved):

PR #353, opened 2022-01-21
Create python-publish.yml

lucpeterson:


.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,36 @@
...
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install build
+    - name: Build package
+      run: python -m build
+    - name: Publish package
+      uses: pypa/gh-action-pypi-publish@27b31702a0e7fc50959f5ad993c78deac1bdfc29
```
koning, line 33: That's a very specific commit, is there a story behind that?

lucpeterson:
This was "auto-generated" when I clicked the GitHub action button for this. It looks like this hash is tied to v.1.4.2 (Feb 2021). They just added a v 1.5.0 a few days ago we could change it to, but this is the script that GitHub writ large creates when you use it to add this action.

.github/workflows/python-publish.yml
lucpeterson: This was "auto-generated" when I clicked the GitHub action button for this. It looks like this hash is tied to v.1.4.2 (Feb 2021). They just added a v 1.5.0 a few days ago we could change it to, but this is the script that GitHub writ large creates when you use it to add this action.

koning: OK, looks good to me.

koning (approved):

PR #359, opened 2022-03-28
Docs update and rename lgtm.yml

koning:


lucpeterson:
Does this have to be into main or is develop a better place?

lucpeterson:
Ok sounds good. The test is failing because there's a black import error? Is this an issue with a library changing on us?

koning:
Yes, I believe so. Is there a way to constrain the python version used for the lint step?

koning:
Looks like the lint worked that time.

lucpeterson (approved):

PR #360, opened 2022-04-09
Added Server Command Feature to Merlin

ryannova:
Added an additional feature to the Merlin. The new server commands allow users to initialize containerized Redis servers, which can be utilized by Merlin as a broker and results server. It uses singularity and pulls the latest Redis docker container off of Docker Hub to create a redis_latest.sif container. 

I have also added a redis.conf file which can be used to properly configure the redis instance. The configuration file currently has most of the default settings with snapshot and append only mode enabled. The appendonly.aof corresponds to the append only file and the dump.rbd corresponds to the snapshots of the redis database.

The following commands have been added.

    merlin server init
    merlin server start
    merlin server stop
    merlin server status

Running `init` creates a merlin_server folder in the current directory that will be used to store the redis containers and all the necessary files to allow merlin server commands to function. It will fetches the latest redis docker container off of Docker Hub and create a redis-latest.sif for singularity to use. Also it copies over the redis.conf file into the merlin_server to configure redis when an instance is ran.

Running `start` will attempt to start the redis-latest singularity container and use the "./merlin_server/redis.conf" file as configuration. `init` takes copies over the default redis config file but it can be configured by the user or a user generated one can be used instead. A "merlin_server.pid" file will also be created to log the pid of the containerized process.

Running `stop` will attempt to read the pid file in merlin_server and send a kill command on the containerized process, and report to the user if it was successful.

Running `status` will let the user know the current status of the containerized redis server. It will let the user know if it has been initialized, if there is a valid container in the merlin_server sub-directory, if the there is a instance of the container that is currently running, or if there is no container processes.

Edits:

- Change to utilize configuration file and parse redis output. [info](https://github.com/LLNL/merlin/pull/360#issuecomment-1104603820)


koning:
This looks good Ryan. Can you also update the docs with these commands and add this method to the server section?

ryannova:
Definitely I'll update docs with these commands into the docs folder of the repo with another commit. There are also some changes that Luc requested, so I will make those changes in other commits as well.

lucpeterson:
I'd like to change this to merge into develop instead of main

koning:
Some other items to ponder are alternative image runtime methods (podman, docker) and alternative server types and configurations and where these overrides might be located.
e.g
~/.merlin/servers/redis.server
container:
    # Select the format of the recipe e.g. singularity, docker, podman
    format: singularity
    # The image name
    image: redis_latest.sif
    # The url to pull the image from
    url: docker://redis
    # The config file
    config: redis.conf
    # Which servers to handle,  [broker, results]  is default
    server: [broker, results]
    # Which machines to run the server on
    # machines: [cori, pellmutter]

or redis.server:
container:
    # Select the format of the recipe e.g. singularity, docker, podman
    format: singularity
    image: redis.sif
    config: redis.conf
    url: docker://redis
    server: [results]
    # machines: [cori, pellmutter]
rabbit.server
container:
    format: singularity
    image: rabbitmq.sif
    url: docker://rabbitmq
    config: rabbitmq.conf
    server: [broker]

lucpeterson:
the ability to have a url specified by a file would enable us to use this on air gapped systems with pre-built containers,
eg
container:
    format: singularity
    image: rabbitmq.sif
    url: file://path/to/prebuilt/containers

ryannova:
I have pushed some commits which creates a configuration server directory into ".merlin/". The values are not currently used but I am working on implementing those configuration into the server commands.

koning:
Another configuration suggestion is to separate the container commands from the python code so they can be edited/overridden.

singularity.yaml
```
singularity:
    command: singularity
    init_command: <command> .. (optional or default)
    start_command: <command> run <image>
    stop_command: <command> (optional or kill default)
    pull_command: <command> pull <image> <url>
    ...
```

server/redis_server.yaml
```diff
@@ -0,0 +1,9 @@
+container:
+    # Select the format for the recipe e.g. singularity, docker, podman (currently singularity is the only working option.)
+    format: singularity
+    # The image name
+    image: redis_latest.sif
+    # The url to pull the image from
+    url: docker://redis
+    # The config file
+    config: redis.conf
```
koning, line 9: The cool git checker is complaining about the lack of a newline here.

ryannova: Added an extra line to the end of the configuration file. I believe it should have fixed the issue.

ryannova:
Regarding the configuration for singularity. I think its a good idea to have configuration files or hard coded values or parameters, but is this something we should allow users to change easily. My concerns is that the format of commands for programs like docker and singularity doesn't seem to change frequently and by allowing users to change command formats it might allow users to break these commands more easily.

koning:
I think we have run into many times that the launcher arguments change slightly because the vendor is working around an issue.  Having the commands external to the python code will allow this change without needing a new merlin version. One example is the in some cases you need to run podman with:
```
srun -n1 --userns podman
```

ryannova:
I see. I'll add three configuration files for singularity, docker, and podman and have them included in the ".merlin/server/" configuration directory. Then the user can use a specific configuration file by changing the format field in the redis_server config file.

server/docker.yaml
```diff
@@ -0,0 +1,6 @@
+docker:
+    command: docker
+    init_command: \{command} .. (optional or default)
+    start_command: \{command} run {image}
+    stop_command: \{command} (optional or kill default)
+    pull_command: \{command} pull {image} {url}
```
koning, line 6: The server yamls need newlines.

server/docker.yaml
```diff
@@ -0,0 +1,6 @@
+docker:
```
koning, line 1: The comment version was just an example. The full version would be something like this:
```
docker:
    command: docker
    # init_command: ?
    start_command: \{command} run --name {name} -d {image}
    stop_command: \{command} stop {name}
    pull_command: \{command} pull {image}

```

You may need to add a name as well to the server config so docker can stop it by name.

server/podman.yaml
```diff
@@ -0,0 +1,6 @@
+podman:
```
koning, line 1: This would be:
```
podman:
   command: podman
    # init_command: \{command} .. (optional or default)
    start_command: \{command} run --name {name} -d {image}
    stop_command: \{command} stop {name}
    pull_command: \{command} pull {image} 

```

server/docker.yaml
ryannova: Added the newlines to the server yamls as well

server/docker.yaml
ryannova: Updated the docker.yaml file

server/podman.yaml
ryannova: Updated the podman.yaml

ryannova:
I made the changes for docker.yaml and podman.yaml, but I think I'm going to hold off on implementing them just yet. I want to try to get everything working well with singularity and then I can add functionality to use docker and podman.

server/singularity.yaml
```diff
@@ -0,0 +1,6 @@
+singularity:
+    command: singularity
+    init_command: \{command} .. (optional or default)
```
koning, line 3: Comment out

server/singularity.yaml
```diff
@@ -0,0 +1,6 @@
...
+    start_command: \{command} run {image}
+    stop_command: \{command} (optional or kill default)
```
koning, line 5: Comment out

koning:
The docker and podman systems don't download the image into a given directory. They have a centralized server location, so you don't need to construct the paths for those.

server/server_setup.py
```diff
@@ -0,0 +1,241 @@
...
+        os.mkdir(server_dir)
+
+    image_path = os.path.join(server_dir, image_name)
+
+    if os.path.exists(image_path):
+        LOG.info(f"{image_path} already exists.")
+        return False
+
+    LOG.info("Fetching redis image from docker://redis.")
+    subprocess.run(["singularity", "pull", image_path, "docker://redis"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
koning, line 137: These should be replaced with a fstring replaced version of the one from the yaml config.

server/server_setup.py
```diff
@@ -0,0 +1,241 @@
...
+        LOG.info("Merlin server has not been initialized. Please run 'merlin server init' first.")
+        return False
+
+    if current_status == ServerStatus.RUNNING:
+        LOG.info("Merlin server already running.")
+        LOG.info("Stop current server with 'merlin server stop' before attempting to start a new server.")
+        return False
+
+    process = subprocess.Popen(
+        ["singularity", "run", os.path.join(server_dir, image_name), os.path.join(server_dir, CONFIG_FILE)],
```
koning, line 167: fstring replace

server/server_setup.py
```diff
@@ -0,0 +1,241 @@
...
+    Stop running merlin server containers.
+
+    :param `server_dir`: location of all server related files.
+    :param `image_name`: name of the image when fetched.
+    """
+    if get_server_status(server_dir=server_dir, image_name=image_name) != ServerStatus.RUNNING:
+        LOG.info("There is no instance of merlin server running.")
+        LOG.info("Start a merlin server first with 'merlin server start'")
+        return False
+
```
koning, line 197: Check for a non-empty stop_command and kill if not present.

server/singularity.yaml
ryannova: init_command is commented out for singularity

server/singularity.yaml
ryannova: commented

server/server_setup.py
ryannova: replaced with yaml config file

server/server_setup.py
ryannova: replaced with fstring from config file

server/server_setup.py
ryannova: Checks have been added to ensure that necessary command values are defined in the config file.

ryannova:
Made notable changes to functions of the merlin server. Now merlin server will utilize configuration files in yaml format for its process calls and given informational outputs regarding the containers that it is running.

Firstly functionality has been added to parse configuration files such as merlin_sever.yaml, singularity.yaml, docker.yaml, and podman.yaml. The main merlin_server.yaml configuration file container main configuration data such as the format, image, and configuration used for running the containers. Other configuration files such as that of singularity.yaml, docker.yaml, and podman.yaml contains format specific commands to interact with them. The parsing of the configuration comes with checks to ensure that the necessary key and values are present in the configuration. If there are some missing values or the file is unable to be parsed, an error will be thrown. For non-essential values a default value will be used if a value is not provided, such as save locations and config locations. Configuration files will be located in 1 of 2 places either in the main merlin configuration directory "~/.merlin/server/" or the "merlin_server" subdirectory by default.

Next all subprocess calls will utilizes fstrings, which are taken from the configuration files stated above. All arguments for the fstrings are taken from the same or their corresponding configuration files. 

Lastly the pid file has been changed to a process file(.pf) which allows the code to manage more than just the pid of the process. The file will include important information such as the "parent_pid", "image_pid", "port", "hostname" and etc. Values coming from the container such as the port is parsed from the output when it is ran. Other information such as process are outputs from subprocess calls and hostnames are outputs from the socket library. An output of the hostname and port is made to the log when an instance of the container is ran successfully.

merlin/server/server_setup.py
```diff
@@ -0,0 +1,291 @@
...
+
+LOG = logging.getLogger("merlin")
+
+
+class ServerStatus(enum.Enum):
+    """
+    Different states in which the server can be in.
+    """
+
+    NOT_INITALIZED = 0
```
lucpeterson, line 38: Maybe have RUNNING as state 0, consistent with the notion that 0 is usually no-error. The others seem like errors. (I know their Eums and not really integers, so maybe it doesn't matter). Thoughts @koning ?

lucpeterson (changes requested):
My only big request is to add some tests to the integration test suite, so that we know this capability works as we do PRs, merges, etc. We'll want to make sure this is tested.

merlin/server/server_setup.py
koning: Yes, nonzero is usually a error. RUNNING as state 0 is good with the default state set as NOT_RUNNING.

koning:
Another thing to consider is a single user starting different servers for different studies and where the input/ouput will go in that case.

merlin/server/server_config.py
```diff
@@ -0,0 +1,105 @@
+import logging
+import os
+
+import yaml
+
+
+LOG = logging.getLogger("merlin")
+
+MERLIN_CONFIG_DIR = os.path.expanduser("~") + "/.merlin/"
```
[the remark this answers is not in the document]
ryannova: These values are defined in [configfile.py](https://github.com/LLNL/merlin/blob/0497a98f5d29ef0047b1db39ff362696df7d79e5/merlin/config/configfile.py#L50-L51), but I'm not able to import this value since there is a [get_config()](https://github.com/LLNL/merlin/blob/0497a98f5d29ef0047b1db39ff362696df7d79e5/merlin/config/configfile.py#L305-L306) call at the end of configfile.py, which requires the ~/.merlin directory to exist unless it will error. I noticed that in other places the configuration directory is locally defined such as here in the [main.py](https://github.com/LLNL/merlin/blob/0497a98f5d29ef0047b1db39ff362696df7d79e5/merlin/main.py#L312-L313) file, so I defined it globally in the server_config.py file.

merlin/server/server_setup.py
ryannova: Changed the RUNNING state as 0

ryannova:
> My only big request is to add some tests to the integration test suite, so that we know this capability works as we do PRs, merges, etc. We'll want to make sure this is tested.

Added tests into the integration test suite to test for the initialization of the merlin server command and also start and stop commands.

ryannova:
> Another thing to consider is a single user starting different servers for different studies and where the input/ouput will go in that case.

The merlin server functionally creates the image and container configuration files locally into a "merlin_server/"(default name) subdirectory based on where the user runs the "merlin server init" command. So if a user is trying to run multiple servers for different studies they can create different directory to store the input/output.

For example if the user has two studies they can create two directories "study1/" and "study2/". Within "study1/" directory a user will run "merlin server init" which creates a subdirectory "study1/merlin_server/". This is repeated in "study2/" directory. Now the user can manage and monitor each container by simply running the merlin server command in the corresponding directory to the study.
The directory structure will be
- study1/
  - merlin_server/
    - [configuration files]
  - [input and output files]
- study2/
  - merlin_server/
    - [configuration files]
  - [input and output files]

docs/source/merlin_server.rst
```diff
@@ -0,0 +1,70 @@
+Merlin Server
+=============
+The merlin server command allows users easy access to containerized broker 
+and results servers for merlin workflows. The main configuration will be
+stored in the subdirectory called "server/" by default in the main merlin
+configuration "~/.merlin". However different server images can be created
+for different use cases or studies just by simply created a new directory
```
lucpeterson, line 7: "by simply creating"

docs/source/merlin_server.rst
```diff
@@ -0,0 +1,70 @@
...
+
+First create and navigate into a directory to store your local merlin 
+configuration for a specific use case or study.
+
+.. code-block:: bash
+
+   mkdir study1/
+   cd study1/
+
+After wards you can instantiate merlin server in this directory by running
```
lucpeterson, line 20: Afterwards

docs/source/merlin_server.rst
```diff
@@ -0,0 +1,70 @@
+Merlin Server
+=============
+The merlin server command allows users easy access to containerized broker 
+and results servers for merlin workflows. The main configuration will be
```
lucpeterson, line 4: Add a sentence after the first:
"This allows users to run merlin without a dedicated external server."
Then a blank line to have a new paragraph before "The main configuration"

docs/source/merlin_server.rst
```diff
@@ -0,0 +1,70 @@
...
+   ~/study1$ ls merlin_server/
+   redis.conf  redis_latest.sif
+
+The main configuration in "~/.merlin/server" deals with defaults and 
+technical commands that might be used for setting up the merlin server
+local configuration and its containers. Each container has their own
+configuration file to allow users to be able to switch between different
+containerized services freely.
+
+The local configuration "merlin_server" contains configuration specific to
```
lucpeterson, line 48: I think this needs to be rewritten. Perhaps

The local configuration "merlin_server" **folder** contains configuration **files** specific to a certain use case or run.

docs/source/merlin_server.rst
```diff
@@ -0,0 +1,70 @@
...
+a certain use case or run. In the case above you can see that we have a 
+redis singularity container called "redis_latest.sif" with the redis configuration
+file called "redis.conf". This redis configuration will allow the user to
+configurate redis to their specified needs without have to manage or edit
+the redis container. When the server is ran this configuration will be dynamically
```
lucpeterson, line 53: "the server is run"

docs/source/merlin_server.rst
```diff
@@ -0,0 +1,70 @@
...
+read, so settings can be changed between runs if needed.
+
+Once merlin server has been initialized in the local directory the user will be allowed
```
lucpeterson, line 56: Once the merlin server

docs/source/merlin_server.rst
```diff
@@ -0,0 +1,70 @@
...
+to run other merlin server commands such as "run, status, stop" to interact with the 
+merlin server. A detail list of commands can be found in the `Merlin Server Commands <modules/server/commands.html>`_ page.
```
lucpeterson, line 58: Also is that the correct way to reference another page? I think this is correct too: `Merlin Server Commands <./modules/server/commands>`

docs/source/merlin_server.rst
```diff
@@ -0,0 +1,70 @@
...
+
+Note: Running "merlin server init" again will NOT override any exisiting configuration
+that the users might have set or edited. By running this command again any missing files 
+will be created for the users with exisiting defaults. HOWEVER it is highly advised that 
+users back up their configruation in case an error occurs where configuration files are override.
```
lucpeterson, line 63: users back up their _configuration_ in case an error occurs and configuration files are unintentionally overridden.

docs/source/merlin_server.rst
lucpeterson: Also the "modules" subdirectory currently houses modules for the tutorial. I don't think we should add these documents to that subfolder

docs/source/modules/server/commands.rst
```diff
@@ -0,0 +1,40 @@
...
+Merlin server has a list of commands for interacting with the broker and results server.
+These commands allow the user to manage and monitor the exisiting server and create 
+instances of servers if needed.
+
+Initializing Merlin Server (``merlin server init``)
+---------------------------------------------------
+The merlin server init command creates configurations for merlin server commands.
+
+A main merlin sever configuration subdirectory is created in "~/.merlin/server" which contains 
+configuration for local merlin configuration, and configruations for different containerized
```
lucpeterson, line 13: typo: "configurations"

docs/source/modules/server/commands.rst
```diff
@@ -0,0 +1,40 @@
...
+services that merlin server supports, which includes singularity (docker and podman implemented
+in the future). 
+
+A local merlin server configuration subdirectory called "merlin_server/" will also
+be created when this command is ran. This will contain a container for merlin server and associated
```
lucpeterson, line 18: when this command is run

docs/source/modules/server/commands.rst
```diff
@@ -0,0 +1,40 @@
...
+configuration files that might be used to start the server. For example, for a redis server a "redis.conf"
+will contain settings which will be dynamically loaded when the redis server is ran. This local configuration
```
lucpeterson, line 20: when the redis server is run

docs/source/modules/server/commands.rst
```diff
@@ -0,0 +1,40 @@
...
+will also contain information of currently running containers as well.
+
+Note: If there is an exisiting subdirectory containing merlin server configuration then only
```
lucpeterson, line 23: containing _a_ merlin server configuration

tests/integration/test_definitions.py
```diff
@@ -45,6 +45,19 @@ def define_tests():
...
         ),
     }
+    server_tests = {
+        "merlin server init": ("merlin server init", HasRegex(".*successful"), "local"),
+        "merlin server start/stop": (
+            "merlin server start; merlin server status; merlin server stop",
+            [
+                HasRegex("Server started with PID [0-9]*"),
+                HasRegex("Merlin server is running"),
+                HasRegex("Merlin server terminated"),
```
lucpeterson, line 55: Good! Once we have an easy configuration, it'd be good to have a test that confirms that Merlin can connect to the right server, eg that "merlin info" returns ok

lucpeterson (changes requested):
Good work. I pointed out some grammar/typo issues in the docs. Also a request that we move the files out of the "modules" subdirectory, which is intended for the tutorial modules

docs/source/merlin_server.rst
ryannova: changes made

docs/source/merlin_server.rst
ryannova: changes made

docs/source/merlin_server.rst
ryannova: changes made

docs/source/merlin_server.rst
ryannova: changes made

docs/source/merlin_server.rst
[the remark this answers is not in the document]
ryannova: changes made

ryannova: I moved it out of the modules subdirectory and placed the server directory directly in the source directory.

docs/source/modules/server/commands.rst
```diff
@@ -0,0 +1,40 @@
+Merlin Server Commands
```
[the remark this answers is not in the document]
ryannova: server docs directory moved into source directory

docs/source/modules/server/commands.rst
ryannova: changes made

docs/source/modules/server/commands.rst
ryannova: changes made

docs/source/modules/server/commands.rst
ryannova: changes made

docs/source/modules/server/commands.rst
```diff
@@ -0,0 +1,40 @@
...
+A main merlin sever configuration subdirectory is created in "~/.merlin/server" which contains 
+configuration for local merlin configuration, and configruations for different containerized
+services that merlin server supports, which includes singularity (docker and podman implemented
+in the future). 
+
+A local merlin server configuration subdirectory called "merlin_server/" will also
+be created when this command is ran. This will contain a container for merlin server and associated
+configuration files that might be used to start the server. For example, for a redis server a "redis.conf"
+will contain settings which will be dynamically loaded when the redis server is ran. This local configuration
+will also contain information of currently running containers as well.
```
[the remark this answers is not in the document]
ryannova: changes made

docs/source/modules/server/commands.rst
ryannova: changes made

tests/integration/test_definitions.py
ryannova: Agreed, something that might be useful is for a command or an easy interface where users will be able to switch freely between using the merlin server and using an external server like before. The only way that I have been able to do it so far is to change the app.yaml file manually within ~/.merlin/. It might be good to have something like "merlin server set local" and "merlin server set remote" to allow users to switch between the two without having to manually change the file.

ryannova:
Thanks for catching to grammer/typo issues. I have made the corrections in the docs. I have also moved the server directory out of the modules subdirectory and into the source directory.

lucpeterson (approved):

tests/integration/test_definitions.py
lucpeterson: Let's make an issue on this

lucpeterson:
@koning I'm good with this. Once you approve, I'll merge into develop

koning (approved):

lucpeterson (changes requested):
I just noticed that the new tests actually FAILED, but the test seems to have registered as passed. This shows us two major bugs we need to deal with
1. The tests should pass
2. If a test fails, the GitHub action should report the failure as well

lucpeterson:
> I just noticed that the new tests actually FAILED, but the test seems to have registered as passed. This shows us two major bugs we need to deal with
> 
> 1. The tests should pass
> 2. If a test fails, the GitHub action should report the failure as well

@koning it looks like the lgtm change to run_tests.py removed the sys.exit(main()) call, which broke the integration tests. main returns a non-zero exit status, but the run_tests.py command doesn't do anything with that return value, so it exits normally and the test is just happy

ryannova:
Thanks for catching that when I was reviewing the code I noticed the same missing exit() call. I went and updated the code in run_tests.py to include an exit() call on the return value of main().

lucpeterson:
> Thanks for catching that when I was reviewing the code I noticed the same missing exit() call. I went and updated the code in run_tests.py to include an exit() call on the return value of main().

I'd like @koning to weigh in here. We made this change because the lgtm CI said this was an error, but I'm not sure if this is the "right" way to do it, or if there's another way

koning:
I believe the lgtm comment was that the sys.exit had no effect, I think we should return a value instead of calling sys.exit().

lucpeterson:
https://lgtm.com/rules/3960095/

ryannova:
I added changes to the workflow to fix the issue with the merlin server tests failing. The containers were missing singularity to run the tests properly, so I added a setup step to install singularity in the workflow.

lucpeterson (approved):
I'm good with this now and will merge it in.
@koning we reverted the sys.exit(main()) to what it was before, knowing that it'll throw an lgtm warning. Let's see exactly what it is and then come up with a solution that makes it happy in a new PR. (We also have a lot of linting we could do anyway)

PR #363, opened 2022-04-28
Fix lgtm returns

koning:


lucpeterson:
i merged in the other PR first. need update this from the new develop?

lucpeterson:
Is this how it wants it done? Everything in try-catch? Seems kind of heavy handed

koning:
I agree, that is what lgtm suggested. The issue is the default return for a function is None. But if the tests are looking for a nonzero exit code, they need to be provided. The original sys.exit(main()) was just returning zero always except for the two files you changed.

lucpeterson (approved):

PR #368, opened 2022-06-16
Addition of new shortcuts in specification file

jimagaffney:
Adding three new shortcuts to the specification definition:

`$(MERLIN_SAMPLE_VECTOR)` is substituted for `"$(SAMPLE_COLUMN_1) $(SAMPLE_COLUMN_2) ..."`
`$(MERLIN_SAMPLE_NAMES)` is substituted for `"SAMPLE_COLUMN_1 SAMPLE_COLUMN_2 ..."`
`$(MERLIN_SPEC)` is substituted for the path to the `*.expanded.yaml` file in the `merlin_info` directory

Changelog:

- In the `special_vars` method of the `MerlinStudy` class, add reference and definitions of the new shortcuts
- Added new method `get_sample_labels` to `MerlinStudy` class. This generalizes the existing `sample_labels` property to allow the labels to be read from a user-specified `MerlinSpec` instance. This is needed to avoid access to the `expanded_spec` property before `special_vars` are defined, which causes an `AttributeError`
- Update to existing `sample_labels` property to be consistent with new `get_sample_labels` method

Tests Performed:

- Successful generation and run of "hello world" example script
- Successful generate and run of existing distributed workflow modified to use new shortcuts

### DIFF
--- a/merlin/study/study.py
+++ b/merlin/study/study.py
@@ -107,6 +107,11 @@ class MerlinStudy:
             "MERLIN_SOFT_FAIL": str(int(ReturnCode.SOFT_FAIL)),
             "MERLIN_HARD_FAIL": str(int(ReturnCode.HARD_FAIL)),
             "MERLIN_RETRY": str(int(ReturnCode.RETRY)),
+            "MERLIN_ALL_SAMPLES": " ".join(self.get_sample_labels(from_spec=self.original_spec)),
+            "MERLIN_SPEC": os.path.join(
+                    self.info,
+                    self.original_spec.description["name"].replace(" ", "_") + ".expanded.yaml"
+            )
         }
 
         self.pgen_file = pgen_file
@@ -182,6 +187,11 @@ class MerlinStudy:
             return self.load_samples()
         return []
 
+    def get_sample_labels(self, from_spec):
+        if from_spec.merlin["samples"]:
+            return from_spec.merlin["samples"]["column_labels"]
+        return []
+
     @property
     def sample_labels(self):
         """
@@ -197,9 +207,7 @@ class MerlinStudy:
 
         :return: list of labels (e.g. ["X0", "X1"] )
         """
-        if self.expanded_spec.merlin["samples"]:
-            return self.expanded_spec.merlin["samples"]["column_labels"]
-        return []
+        return self.get_sample_labels(from_spec=self.expanded_spec)
 
     def load_samples(self):
         """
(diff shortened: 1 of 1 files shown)


lucpeterson:
- [ ]  I'd like to change the spec shortcut name (and add a few others). What about these names?

- $(MERLIN_SPEC_RAW) --> *.orig.yaml
- $(MERLIN_SPEC_RUN) --> *.partial.yaml
- $(MERLIN_SPEC_EXPANDED) --> *.expanded.yaml

- [ ] We should also add these to the documentation
- [ ] Lint check is failing

merlin/study/study.py
```diff
@@ -107,6 +107,13 @@ def __init__(
             "MERLIN_SOFT_FAIL": str(int(ReturnCode.SOFT_FAIL)),
             "MERLIN_HARD_FAIL": str(int(ReturnCode.HARD_FAIL)),
             "MERLIN_RETRY": str(int(ReturnCode.RETRY)),
+            # below will be substituted for sample values on execution
+            "MERLIN_SAMPLE_VECTOR": " ".join([ "$({})".format(k) for k in self.get_sample_labels(from_spec=self.original_spec)]),
+            "MERLIN_SAMPLE_NAMES": " ".join(self.get_sample_labels(from_spec=self.original_spec)),
+            "MERLIN_SPEC": os.path.join(
```
koning, line 113: These are good additions, I think we should add options for all three spec files in the

koning: The expanded yaml is the one that is actually RUN, maybe RUN should be changed to PARTIAL.

lucpeterson: How about
MERLIN_SPEC_ORIGINAL
MERLIN_SPEC_REQUESTED
MERLIN_SPEC_RUN
??? Other votes? not sure here

ymubarka: I vote for
MERLIN_SPEC_ORIGINAL
MERLIN_SPEC_PARTIAL
MERLIN_SPEC_RUN

so that the names are mostly similar.

lucpeterson: Yeah I don't really like "partial" b/c it's not very descriptive. It's really the original with any command-line override variables substituted. I can't think of a good word tho.

koning: NOSAMPLES ?

ymubarka: EXPANDED_VARS
EXPANDED_ALL?

lucpeterson: Ok, I've (ok @ryannova ) have decided on
MERLIN_SPEC_ORIGINAL_TEMPLATE
MERLIN_SPEC_EXECUTED_RUN
MERLIN_SPEC_ARCHIVED_COPY

koning: LGTM

lucpeterson: agreed. I put @ryannova on it

ryannova:
Resolved by merge request #375

PR #369, opened 2022-07-13
App.yaml for merlin server

ryannova:
Added a new capability to interface with the main app.yaml, and create a new app.yaml containing new configurations specific to merlin server. This should allow users to be able to use the new app.yaml to connect to the merlin server container without having to edit the main configuration in ~/.merlin/. To use a local copy of app.yaml instead of the main configuration there should an app.yaml file in the current run directory. The app.yaml for merlin server is stored in ./merlin_server/ subdirectory, so users will need to move the app.yaml from ./merlin_server/ to the current running directory.


merlin/server/server_util.py
```diff
@@ -508,3 +514,38 @@ def apply_to_redis(self, host: str, port: int, password: str) -> None:
         for user in current_users:
             if user not in self.users:
                 db.acl_deluser(user)
+
+
+class AppYaml:
```
lucpeterson, line 519: it seems like this class could have some overlap / utility for the configfile part of the code. @koning , thoughts about how we could use this to refactor?

koning: Most of the server config can come from the app.yaml (username, passwords, port , server_type). we may be able to just add the rest of the container config to the server (broker , results_backend) sections.

ryannova: The container config corresponds to the [merlin_server.yaml](https://github.com/LLNL/merlin/blob/develop/merlin/server/merlin_server.yaml) config file where it details all of the files and configurations needed to be changed for the server container.

lucpeterson: Ok let's write down an issue to see how we can unify this with the configfile code

lucpeterson (approved):

PR #370, opened 2022-07-21
Bugfix/changelog ci

lucpeterson:
This checks the changelog against the branch we're merging, not just main


ryannova (approved):
It looks good to me.

Also the shorten version of the changes for merlin server command is in #369

koning (approved):

PR #372, opened 2022-08-08
Updated merlin server unit testing

ryannova:
Updated the tests for the merlin server unit tests to include tests for the latest addition of the merlin server config command. 

In order to properly test the command I changed the testing structure slightly to allow for ending commands, where a command would run(usually for clean up) after the main command and test cases have been ran.


tests/integration/conditions.py
```diff
@@ -238,3 +238,56 @@ def passes(self):
         if self.negate:
             return not self.is_within()
         return self.is_within()
+
+
+class PathExists(Condition):
```
lucpeterson, line 243: the other Condition classes have a __str__ value defined as well, which is called when the test fails, for use in debugging error failures. (ie if not Condition.passes: print(Condition) ) Consider adding this debug message to these classes as well.

tests/integration/test_definitions.py
```diff
@@ -45,18 +54,80 @@ def define_tests():
...
+            merlin server start;
+            merlin server config --remove-user new_user;
+            merlin server stop;
+            """,
+            [
+                FileHasRegex("./merlin_server/redis.users_new", "new_user"),
+                FileHasNoRegex("./merlin_server/redis.users", "new_user"),
+            ],
+            "local",
+            """rm -rf appendonly.aof dump.rdb merlin_server/""",
```
lucpeterson, line 129: This command is called a lot. Is it worth storing this as a variable?

lucpeterson (changes requested):
This looks good. My only request is to add the __str__ property to the new tests

tests/integration/run_tests.py
```diff
@@ -139,7 +145,7 @@ def run_tests(args, tests):
         n_to_run = 0
         selective = True
         for test_id, test in enumerate(tests.values()):
-            if len(test) == 3 and test[2] == "local":
+            if len(test) >= 3 and test[2] == "local":
```
jwhite242, line 148: What's the significance of these magic constants, and what is the conditional here trying to do/enforce?

tests/integration/run_tests.py
```diff
@@ -139,7 +145,7 @@ def run_tests(args, tests):
...
                 args.ids.append(test_id + 1)
```
jwhite242, line 149: Why the need to add the +1 offest to the ids here?

tests/integration/conditions.py
jwhite242: `__repr__` can be another useful attribute for debug output and it + the `__str__` can allow verbose and brief output modes.

tests/integration/test_definitions.py
```diff
@@ -384,7 +455,8 @@ def define_tests():
     all_tests = {}
     for test_dict in [
         basic_checks,
-        server_tests,
+        server_basic_tests,
```
jwhite242, line 458: So rather than re-enumerating things here, what about some alternate patterns like just adding the contents of these sub-dicts to all_tests directly when creating them, or even just having this function be a generator that yields the key: value pairs and a little helper outside that consumes it.  It looks like the names of these test group dicts are used more for documentation than anything, suggesting they don't really need to be directly created at all.

Those are more just some style choices than anything though; my main concern with the design of this (and the above refactors would hit this too as it'd still largely be the same design) is that there's a lot of silent failures/bugs that this dictionary type structure is susceptible to: update and direct key assignment ('dict[key]=newvalue') will happily overwrite the values of any keys that are already there leading to incorrect tests being run if developers aren't careful to scan all of the test names in here and avoid them.  Some error checking here would be great to at least warn of that so such collisions can be more easily traced/fixed.

tests/integration/test_definitions.py
```diff
@@ -45,18 +54,80 @@ def define_tests():
...
+    server_basic_tests = {
+        "merlin server init": (
+            "merlin server init",
+            HasRegex(".*successful"),
+            "local",
+            "rm -rf appendonly.aof dump.rdb merlin_server/",
+        ),
         "merlin server start/stop": (
-            "merlin server start; merlin server status; merlin server stop",
+            """merlin server init;
```
jwhite242, line 65: Pardon my ignorance of some of the larger machinery in here, but what do the ';'s add here that isn't already accomplished by the new lines embedded in those multi-line strings (I'm assuming this is a splitting mechanism to enable accessing the individual commands?)

jwhite242 (commented):
Mostly just comments/questions.  Most important changes I'd really request for this PR are the error checking/logging on the test definitions to catch collisions.

tests/integration/conditions.py
ryannova: Thanks for catching that I didn't noticed that the `__str__` was used for debugging and error reporting. I have added the __str__ but omitted the __repr__ because I don't think it is currently used. I want to keep it consistent with the other existing conditions.

tests/integration/test_definitions.py
ryannova: The command is used for cleaning the files directories created by merlin server and this ensures that new merlin server tests starts new without any artifacts. There might be some changes to this command if configurations are changed, but I agree with you that it might be worthwhile to have a variable for this command. I have added one as a constant at the top.

tests/integration/run_tests.py
```diff
@@ -78,6 +78,12 @@ def run_single_test(name, test, test_label="", buffer_length=50):
             info["violated_condition"] = (condition, i, len(conditions))
             break
 
+    if len(test) == 4:
```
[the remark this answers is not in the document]
ryannova: Thanks for pointing this out. These magic constants/indexes are for accessing the tuple that are defined in test_definition.py. I'm not sure about the original reason for implementing the testing structure this way, but maybe @lucpeterson and @koning can provide better insight on this. While I agree that this isn't a great way of doing it because it requires refactoring of all instances of test[] when new features needs to be added, it will take a decent amount of time and effort to change the current implementation. I am not sure if we have allocated time in our schedule to do that and this pull request is trying to accommodate for some of the requirements for the "merlin server" command, and properly test it.

tests/integration/run_tests.py
ryannova: This is to accommodate for the tuple increasing in size. The variable "local" is located in the 3rd element of the tuple and thus require the tuple to be at least of size 3. Regarding magic constants please see my comment above.

tests/integration/run_tests.py
ryannova: Please see my comment above. It is to accommodate for the increase in the tuple size.

tests/integration/test_definitions.py
ryannova: I agree with you that this is following the same design structure as the testing structure. From what I can see you are right that the names of the test are mainly for documentation. See my comment above, I am not sure about the original reason for making it this way but in order to change the design of the testing structure we would need to refactor and redesign the framework

tests/integration/test_definitions.py
ryannova: The addition of ';'s is because the command are all run with a shell. The testing framework doesn't parse the commands but runs the string with subprocess.Popen() with shell. With most shell-scripts ";" is to designate the end of a command. While spaces and new-lines are used for separating arguments of a given command.

tests/integration/run_tests.py
jwhite242: Sure, but why does the tuple need to be increased.  What's the requirement driving this downstream?  and can there be some comments in here explaining the why so that reverse engineering of the whole test call stack isn't needed to figure it out.  At first glance from a new dev who's unfamiliar with this code it's just unclear why this conditional is here and also why the '>=' is the correct  condition.

tests/integration/test_definitions.py
jwhite242: Yeah, that's fair on the refactoring.   I'd still recommend some error checking to catch the silent failures that can happen here when joining all these dictionaries together

tests/integration/run_tests.py
ryannova: The increase in the tuple size is needed for a command to be ran after the completion of the unit test. This ending command is located at the 4th element of the tuple. It definitely possible to encode that information within the first element of the tuple, but it will require parsing and separating of the initial commands, and the ending commands later downstream.  This is needed specifically for `merlin server` command testing because it creates many different directories and files to stores config information. Without someway of cleaning up these configurations any subsequent `merlin server` commands will be using previous existing configurations. You are right that it would be great to have an additional comment would be great for future devs, so I have add one to explain the reason for the condition.

jwhite242: Great, thanks for the explanation and the comment!

tests/integration/run_tests.py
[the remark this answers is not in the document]
jwhite242: Similar to the other thread you might put a short comment on some of them -> i.e. this particular condition handles the non-local cases that have some extra cleanup commands attached to them based on your other explanation?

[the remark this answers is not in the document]
koning: I thin Ben Bay wrote the tests, I don't recall the reason for the indexing.

lucpeterson (approved):

tests/integration/run_tests.py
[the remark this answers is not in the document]
lucpeterson: Let's decide to clean this up in another PR, that for instance turns the indexed tuple into some simple object with explicit "pre_command" "test_command" and "post_command" fields or something like that

lucpeterson (approved):

PR #367, opened 2022-06-14
Add support for non-merlin blocks in specification file

jimagaffney:
Adding capability for non-user blocks of yaml code to be included in the merlin specification, and propegated to the .partial and .expanded versions at runtime. This allows the users to define yaml specifications for steps in the workflow in the same file, and for user code to perform consistency checks between the merlin spec and code in the steps - for example, checking consistency between variable names, etc.

A new "user" block has been added to the `MerlinSpec` class in `spec/specification.py`. The `user` block is not included in any checking and does not have default values. The yaml code has shell and merlin parameters substituted and is copied to the working directories verbatim. 

Changelog:

- Added a static method `load_user_block` to the `MerlinSpec` class, which loads the `user` block from the yaml specification and returns the python representation
- In functions `load_specification` and `load_spec_from_string`, added definitions of `self.user` through calls to `load_user_block`
- In the `yaml_sections` and `sections` properties, added references to `self.user`
- In the `dict_to_string` method of `MerlinSpec` object, add `user` block support for pretty printing of lists

When no `user` block is present in the yaml file, the `MerlinSpec.user` attribute is an empty dictionary

Test Functionality:

- correct generation of problems in `merlin` test problems (with `-clean` option)
- correct generation of problems and variable substitution when a `user` block is added to the test problems (with `-clean` option)
- complete execution of `merlin` test problems

PR #376, opened 2022-09-23
Add support for non-merlin blocks in specification file

ryannova:
Updated #367.

Updated:
Updated Changelog


koning:
These PRs need some doc updates.

ryannova:
I'm not too familiar with the merlin workflows yet, so I'm not sure what is the purpose for the addition of non-merlin "user" block. If you could let me know what to add for docs I can include it in the merlin_specification.rst and merlin_variables.rst where it is documented.

koning:
We can workshop it in the comments, maybe @jimagaffney can give us an example.

lucpeterson:
The user block allows other variables / yaml syntax in the workflow file to be propagated through to the workflow (including in the .parital.yaml and .expanded.yaml). This gives the user increased flexibility in how they write their files.

ryannova:
That's great. Is there an sample or example? Since in the current documentation an example was given for each block.

lucpeterson:
> That's great. Is there an sample or example I can use? Since in the current documentation an example was given for each block.

@jimagaffney can you cook up an example?

lucpeterson:
> > That's great. Is there an sample or example I can use? Since in the current documentation an example was given for each block.
> 
> @jimagaffney can you cook up an example?

Jim says he can cook up an example at the end of next week

docs/source/merlin_specification.rst
```diff
@@ -277,6 +277,14 @@ see :doc:`./merlin_variables`.
               batch:
                  type: local
               machines: [host3]
+  ####################################
+  # Merlin Block (Required)
+  ####################################
+  # The user block allows other variables in the workflow file to be propagated
```
koning, line 283: Maybe make a note that this is optional, because it looks like it is required from the comment above.

ryannova: I have updated the documentation on the user block and added the example from the freature_demo to the docs.

lucpeterson:
What's the status of this and the other PR? #378 Are we good?

merlin/examples/workflows/feature_demo/feature_demo.yaml
```diff
@@ -17,92 +17,116 @@ env:
         HELLO: $(SCRIPTS)/hello_world.py
         FEATURES: $(SCRIPTS)/features.json
```
lucpeterson, line 19: i would recommend we just do this for one of the steps instead of all of them

ryannova:
I think we should be good for this one and the other PR. #378 I don't have anything else I need to change for either.

merlin/examples/workflows/hello/hello.yaml
```diff
@@ -10,15 +10,28 @@ global.parameters:
         values : ["world","mundo"]
         label  : WORLD.%%
 
+user:
```
lucpeterson, line 13: If we're changing this, we need to make sure the tutorial aligns with it (since the tutorial uses hello_world). I suggest we don't modify this file and keep it simple

lucpeterson (changes requested):

merlin/examples/workflows/hello/hello.yaml
ryannova: Okay sounds good. I removed it from hello.yaml example

merlin/examples/workflows/feature_demo/feature_demo.yaml
ryannova: Okay. I updated so that it only uses yaml anchors for the first step.

lucpeterson (approved):

koning (approved):

PR #380, opened 2022-10-26
Ssl server check fixes

koning:
This will fix launchit rabbitmq server check issue. The server is actually working, but the check will fail when merlin info is run.


lucpeterson (approved):
good catch

koning:
It was actually Dave Fox who found the issue.

lucpeterson:
Well then Dave get's a cookie!

PR #382, opened 2022-11-01
Fix footer styling in docs

bgunnar5:
Added configuration files including .readthedocs.yaml and docs/requirements.txt to get sphinx updated to the newest version. This fixed the styling issue with the footer and also made docs/source/_static/theme_overrides.css unnecessary so it was removed. Updating sphinx caused issues with the Makefile in docs/ so I modified it by adding a pip install command and a new clean method. 

There was a secondary issue involving the variables page in the docs where the horizontal scroll was overflowing behind the left nav bar. This was also patched by modifying the custom.css file and slightly modifying how the reserved variables table in merlin_variables.rst was formatted.

PR #383, opened 2022-11-07
Docs/install changes

bgunnar5:
I went through all of the docs and added some new instructions where I felt things weren't as clear as they could be. I also added instructions on how to install merlin via make for developers and fixed typos when I found them.

This branch was created off of the fix_footer branch in order to work with the newer version of Sphinx so the commits from Nov 2, 2022 (commit id d10510d) are the same as in the fix_footer branch that's waiting to be merged.


docs/source/merlin_variables.rst
```diff
@@ -31,73 +31,138 @@ The directory structure of merlin output looks like this:
...
+
+            /*/*/*/*
+ 
+   * - ``$(MERLIN_PATHS_ALL)``
+     - A space delimited string of all of the paths;
+       can be used as is in bash for loop for instance with:
+ 
+       .. code-block:: bash
+ 
+          for path in $(MERLIN_PATHS_ALL)
```
lucpeterson, line 115: One of these blocks shouldn't be here

docs/source/modules/hello_world/hello_world.rst
```diff
@@ -234,6 +240,12 @@ Immediately after that, this will pop up:
 .. literalinclude :: celery.txt
    :language: text
 
+You may not see all of the info logs listed after the Celery C is displayed. If you'd like to see them you can change Merlin's log levels with the ``-lvl`` or ``--level`` tag. For more information on this, you can run:
```
lucpeterson, line 243: this is actually changed via the workers' log level, not merlin's

lucpeterson (changes requested):
just a few comments, but this is a huge improvement!

docs/source/modules/hello_world/hello_world.rst
bgunnar5: Do you have an example of how to change their log levels via the CLI? When I run `merlin run-workers --help` the optional arguments don't show a log level flag to set. The only optional flags that are listed for run-workers are: --help, --worker-args, --steps, --echo, and --vars.

bgunnar5: @lucpeterson (not sure if it notifies you when I comment so figured I'd add this @ here)

lucpeterson (approved):
Does this PR make #382 now obsolete?

PR #385, opened 2022-11-17
Update Merlin Server

ryannova:
Fixed and update a number of issues with the merlin server functionality. 

- Update "default" user password to be automatically generated and stored in redis.pass. (Currently same password as the new user created.) (Fixes issues with celery where only user allowed to be authenticated is "default".)
- Changed singularity configuration to set home directory to be the local configuration directory. (This change means that dump and append files will not be stored in the local configuration directory rather than the run directory) (Fixes the issue with unrecognized directories where singularity could not access directories not located in a sub-directory of the user's home directory)
- Updated merlin server to use app.yaml configuration rather than its own configuration file.

Other changes
- Update Makefile target "fix-style" to be consistent with other style targets in Makefile.


lucpeterson (approved):
Modify the CHANGELOG and I'm good with this.
Great job tracking down that bug!

lucpeterson (approved):
👊

merlin/server/merlin_server.yaml
```diff
@@ -9,8 +9,8 @@ container:
     url: docker://redis
     # The config file
     config: redis.conf
-    # Subdirectory name to store configurations Default: merlin_server/
-    config_dir: merlin_server/
+    # Directory name to store configurations Default: ./merlin_server/
+    config_dir: ./merlin_server/
```
koning, line 13: Where will ./merlin_server/ be by default?

ryannova: Merlin server will still be used for user and container configuration but system configuration will now use app.yaml Before it used a file called merlin_server.yaml that was stored in ~/.merlin/server/.

PR #386, opened 2022-11-21
Update task id for celery tasks

ryannova:
Changed celery tasks id to be its output path rather than a random uuid.


merlin/common/tasks.py
```diff
@@ -96,8 +96,9 @@ def merlin_step(self, *args: Any, **kwargs: Any) -> Optional[ReturnCode]:  # noq
 
     if step:
         self.max_retries = step.max_retries
-        step_name: str = step.name()
-        step_dir: str = step.get_workspace()
+        step_name = step.name()
+        step_dir = step.get_workspace()
+        LOG.debug(self.request.id)
```
lucpeterson, line 101: are these two debug messages giving the same info?

merlin/common/tasks.py
```diff
@@ -281,6 +282,7 @@ def add_merlin_expanded_chain_to_chord(
                     adapter_config=adapter_config,
                 )
                 new_step.set(queue=step.get_task_queue())
+                new_step.set(task_id=os.path.join(workspace, relative_paths[sample_id]))
```
lucpeterson, line 285: are there some tests we can do to make sure this is working correctly?

merlin/common/tasks.py
ryannova: One is displaying the id and the other is displaying the step_name and step_dir but since the dir is the same as the id I think I can remove it.

lucpeterson: yeah. since the step_id is useful to know, maybe combine and say
LOG.debug(f"Running Step {step_name} -- Step ID: {self.request.id}")

lucpeterson: And as I look below at the various other LOG files, I think we should just replace all the directory calls with {self.request.id}
Like LOG.info("fSkipping step {self.request.id}")
I think having the step_name and the step_id in there at the same time could be confusing

merlin/examples/workflows/hello/hello_soft_fail.yaml
```diff
@@ -0,0 +1,24 @@
...
+        label  : GREET.%%
+    WORLD:
+        values : ["world","mundo"]
+        label  : WORLD.%%
+
+study:
+    - name: step_1
+      description: say hello
+      run:
+          cmd: echo "$(GREET), $(WORLD)!"; exit MERLIN_SOFT_FAIL;
```
lucpeterson, line 17: this should be exit $(MERLIN_SOFT_FAIL)

merlin/examples/workflows/hello/hello_soft_fail.yaml
```diff
@@ -0,0 +1,24 @@
+description:
```
lucpeterson, line 1: Can we make a version that is based off hello_samples too? That will test the $(MERLIN_SAMPLE_PATH) hierarchy

lucpeterson: Or maybe a test based off hello_samples?

ryannova: Yeah sure I can add a version for hello_samples.

tests/integration/test_definitions.py
```diff
@@ -127,7 +127,43 @@ def define_tests():
                 FileHasNoRegex("./merlin_server/redis.users", "new_user"),
             ],
             "local",
-            CLEAN_MERLIN_SERVER,
+            "rm -rf merlin_server/ hello*/",
+        ),
+    }
+    server_integrated_tests = {
+        "celery task id": (
+            """merlin server init:
```
lucpeterson, line 135: I think your tests are stuck in limbo because maybe this should be semi-colon?

lucpeterson: I killed the tests after 1 hour

ryannova: I think you are right thanks for catching that. Let me fix it and see if I still stalls.

ryannova:
Branch pushed to feature/task_id. Functionality could be potentially used in `merlin status` command.

PR #389, opened 2022-12-13
Allows loading of nparrays of dtype=object

nkeilbart:
Couldn't load numpy arrays if dtype was set to object. The allow_pickle=True option appears to allow this to happen. Mainly suggesting this change as I couldn't load mixed numpy arrays for my after the sample generation.


lucpeterson:
Update the CHANGELOG and this is good

lucpeterson:
@nkeilbart pull in develop changes so we can see if tests are passing

nkeilbart:
Alright I just merged in the recent updates.

ryannova:
Addressed in #398

PR #393, opened 2022-12-20
Get pip to install merlinspec.json and fix import bug with python 3.7

bgunnar5:
Pip wasn't grabbing merlin/spec/merlinspec.json when you install merlin so I added it to MANIFEST.in. I also came across a bug where "from celery import Celery" was failing in python 3.7. To fix this I had to set the version of importlib_metadata to be below 5.0.0 for python 3.7.

There was yet another bug that came from a numpy release that just happened to come out 2 days ago. The error was saying numpy.str didn't exist. I replaced this with python's built-in str type (safer in the long run I think?).


lucpeterson:
A few things
1. Run make version to bump all the files
2. Is the numpy _str fix backwards compatible?

bgunnar5:
I've created a virtual environment and pip installed v1.8.5 of Merlin. I then ran one of the same tests that was failing because of numpy and it failed again. I'm assuming this means the change is not backwards compatible here? However, if we implement the change to fix this then there are no external changes to our API so I feel like bumping to v2.0.0 seems wrong and there are no features added so bumping to v1.10.0 seems wrong.

I've also considered locking a numpy version in the requirements/release.txt file but that wouldn't fix the issues in previous releases (to my knowledge). Not entirely sure what to do here since I can fix the problem for a new release (v1.9.1) but don't think I'll be able to fix it for previous releases. Any thoughts?

PR #394, opened 2022-12-20
Hotfix for merlin server unable to write config files.

ryannova:
After building the merlin wheel and installing it, users are unable to initiate merlin server.
The cause was due to the fact that the configuration files that lived within the repository couldn't be access and copied over to the users local directory. To circumvent this issue, the problem was fixed by performing the same actions as how the information the "app.yaml" is copied over to the user.

The configuration files are added to merlin as a module and then modules are then loaded, read, and finally written to the user's local directory.


doutriaux1:
@ryannova I hate to be a pain but is there a way to write a test for this? E.g some test that would break before but passes now.

I would like every PR to include a test.

ryannova:
@doutriaux1 There are ways to test for this but not with the current testing structure. All of the current tests are done by cloning the merlin repo and building a virtual environment where tests are then ran using python's subprocess library. To test this you would have to reconfigure github workflow files to build the wheel and install it through that.

doutriaux1:
ok that sounds overkill 😉 let's skip it for this case.

ryannova:
😄 lol okay sounds good. Normally I do try to write a test for each PR, but this was more of a hotfix for features that were working in development environment, but a small discrepancy in the develop and release environment caused it to error during release.
The test for merlin server can be seen [here](https://github.com/LLNL/merlin/blob/b5233f24c7bbde50bbc20e3dbc1194d7abaa3f35/tests/integration/test_definitions.py#L58)

lucpeterson:
@bgunnar5 and I brainstormed a way to modify our testing to test building against the wheel. it should probably be a separate PR tho, since it's a little more involved

doutriaux1:
> 

@lucpeterson  @bgunnar5 I would like to move us to gitlab, could that "future" PR be done on gitlab?

lucpeterson (approved):

PR #395, opened 2022-12-21
Version 1.9.1

ryannova:
- Added merlin/spec/merlinspec.json to MANIFEST.in so pip will actually install it when ran
- Fixed a bug where "from celery import Celery" was failing on python 3.7
- Numpy error about numpy.str not existing from a new numpy release
- Made merlin server configurations into modules that can be loaded and written to users


lucpeterson:
this looks good. question: do we have a test for the .csv and .tab loading that changed? (I don't recall if one of the workflows already checks this)

bgunnar5:
@lucpeterson yeah the "local csv feature_demo" and "local tab feature_demo" tests (tests 33 and 34 I believe) check the csv and tab loading respectively. That's how I found the issue in the first place.

lucpeterson (approved):
Great work. LGTM

lucpeterson:
@bgunnar5  @koning  @doutriaux1 can you take a look at these hot fixes?

bgunnar5 (approved):
LGTM (but I also wrote like half of them)

PR #400, opened 2022-12-23
Feature/openfoam wf update

koning:
Update learn.py to create missing plot
Add singularity base wf.


lucpeterson (commented):
Are there tests we can add for this?

koning:
Do the existing openfoam workflows have a test? Other than running the workflow what would be a good test?

lucpeterson:
@ryannova and @bgunnar5 do you think we could add a test for this new workflow? It uses singularity to run a "real application" in a container and would be great to have that in the test suite. It'd also be nice to have the tutorial use this one (instead of having to build openFoam themselves)

koning:
Two users have been unable to run the tutorial example (openfoam_no_docker) because of the openfoam access issue.

lucpeterson:
Do the changes to the makefile necessitate a change to the isort version? Do we need to change any requirements?

lucpeterson (approved):

PR #404, opened 2023-03-15
Update integration test suite and add stop-workers tests

bgunnar5:
Slightly refactored and added functionality to the integration test suite:
- Instead of using tuples for test definitions they now use dicts
- Added two flags to `run_tests.py`:
  - `--display-tests` to view a table of all the tests and their ids
  - `--distributed` to run only the tests where the `run type` is distributed 
- Added a check to make sure the test definition matches the correct format
- Updated the integration tests' README to accommodate these changes
- Added tests for the stop-workers command (these will fail for now until I update the repo to include the stop-workers fixes)
- Fixed the `negate` option for HasRegex condition


tests/integration/README.md
```diff
@@ -1,21 +1,161 @@
...
+Changing the `cmds` section here to be:
+
+    "cmds": ["echo hello", "echo goodbye"]
+
+would fix this issue and create a valid test definition.
 
 
-### Continuous integration
+## Continuous Integration
 Currently run from [Bamboo](https://lc.llnl.gov/bamboo/chain/admin/config/defaultStages.action?buildKey=MLSI-TES).
```
koning, line 159: This is defunct, bamboo doesn't exist anymore. We should replace this with the github CI info.

tests/integration/conditions.py
```diff
@@ -1,3 +1,33 @@
+###############################################################################
+# Copyright (c) 2022, Lawrence Livermore National Security, LLC.
```
koning, line 2: The year should be updated in all files.

koning (approved):

Issue #396, opened 2022-12-22
"merlin stop-workers" with "--spec" and "--queues" flags does not work


kustowski1:
## Bug Report

"merlin stop-workers" with "--spec" or "--queues" flags does not work

1. Create a simple_spec.yaml, for example: 

description:
  description: Test stop-workers
  name: tmp

study:
- description: my step
  name: my_name
  run:
    cmd: |
       sleep 1000000
    task_queue: my_queue

merlin:
  samples:
    column_labels:
      - junk
    file: $(MERLIN_INFO)/samples.npy
    generate:
      cmd: |-
        spellbook make-samples -dims 1 -n 1 -outfile=$(MERLIN_INFO)/samples.npy

3. merlin --version # my version is 1.8.3
4. merlin run-workers simple_spec.yaml 
5. merlin stop-workers --spec simple_spec.yaml
6. merlin stop-workers --queues my_queue
7. merlin status simple_spec.yaml # this shows that the worker still exists!!!!
8. merlin stop-workers 
9. merlin status simple_spec.yaml # this shows that the worker has been stopped 

**Expected behavior**
I expect to be able to stop workers for a specific spec or queue. merlin stop-workers (without any flags) works but it stops all merlin workers, even from queues and specs that I do not want to stop.

**Please answer these questions to help us pinpoint the problem**
- Does the problem occur in `merlin run --local` mode, distributed mode or neither? 
- -- The problem is related to workers
- If a distributed problem, which backend and queue servers are you using? How are they configured?
- -- rabbit and reddis
- On what machines/architectures are you running merlin? Is this bug on a specific machine or can you reproduce it elsewhere?
- --Can reproduce on multiple machines on LLNL Linux clusters.

**Please run `merlin info` and paste the results here:**

**Additional context**

Issue #401, opened 2023-02-21
[BUG] JSON Schema Validation causes failures with restarts/resubmits


bgunnar5:
## Bug Report

**Describe the bug**
When restarting a job, the yaml spec fails the JSON Schema validation. Restarting a job uses the expanded yaml already created and since the expanded yaml doesn't have arrays in quotes, it fails the validation process.

The same goes for the partial yaml file when you try to resubmit a workflow in an iterative fashion.

**Additional context**
This bug was found by Ryan Tran. If you need more details, reach out to him.

PR #405, opened 2023-03-17
fix bugs for stop-workers and schema validation

bgunnar5:
This branch fixes Issues https://github.com/LLNL/merlin/issues/396 and https://github.com/LLNL/merlin/issues/401

This branch fixes bugs with the `--spec`, `--queues`, and `--workers` flags for the `stop-workers` command. To do this, I had to reformat how workers are launched.


merlin/main.py
```diff
@@ -755,6 +763,13 @@ def generate_worker_touching_parsers(subparsers: ArgumentParser) -> None:
         help="Specify desired Merlin variable values to override those found in the specification. Space-delimited. "
         "Example: '--vars LEARN=path/to/new_learn.py EPOCHS=3'",
     )
+    run_workers.add_argument(
+        "--disable-logs",
+        action="store_true",
+        default=False,
```
lucpeterson, line 769: Remove the default=False line.  It’s not necessary and could perhaps lead to bugs:
https://stackoverflow.com/questions/70428108/whats-the-point-of-having-both-action-store-true-and-default-false-in-parser

merlin/router.py
```diff
@@ -103,12 +104,13 @@ def purge_tasks(task_server, spec, force, steps):
...
     LOG.info(f"Purging queues for steps = {steps}")
 
-    if task_server == "celery":
+    if task_server == "celery":  # pylint: disable=R1705
         queues = spec.make_queue_string(steps)
         # Purge tasks
         return purge_celery_tasks(queues, force)
     else:
         LOG.error("Celery is not specified as the task server!")
+        return -1
```
lucpeterson, line 113: Why the -1? Is this consistent with what we do elsewhere? Should we pin a to-do if creating an enum of return codes?

merlin/spec/specification.py
```diff
@@ -354,8 +374,18 @@ def process_spec_defaults(self):
         if self.merlin["resources"]["workers"] is None:
             self.merlin["resources"]["workers"] = {"default_worker": defaults.WORKER}
         else:
-            for worker, vals in self.merlin["resources"]["workers"].items():
+            steps = self.get_study_step_names()
+            modified_steps = False
+            for _, vals in self.merlin["resources"]["workers"].items():
                 MerlinSpec.fill_missing_defaults(vals, defaults.WORKER)
+                for step in vals["steps"]:
```
lucpeterson, line 381: I’m finding this logic a little hard to follow. Maybe some comments could help?

merlin/study/celeryadapter.py
```diff
@@ -178,10 +178,84 @@ def get_workers_from_app():
...
+    """
+    Helper function to handle creating the kwargs dict that
+    we'll pass to subprocess.Popen when we launch the worker.
+
+    :param `spec`: A MerlinSpec object
+    :returns: A tuple where the first entry is the kwargs and
+              the second entry is variables defined in the spec
+    """
+    senv = spec.environment
+    spenv = os.environ.copy()
```
lucpeterson, line 215: Could we use more descriptive names for senv, spenv, and yenv?
(and yes I know these are legacy variables cut from elsewhere, but should use this opportunity to clean them up a bit)

merlin/study/celeryadapter.py
```diff
@@ -402,14 +481,19 @@ def stop_celery_workers(queues=None, spec_worker_names=None, worker_regex=None):
...
+    from merlin.celery import app  # pylint: disable=C0415
+    from merlin.config.configfile import CONFIG  # pylint: disable=C0415
 
     LOG.debug(f"Sending stop to queues: {queues}, worker_regex: {worker_regex}, spec_worker_names: {spec_worker_names}")
     active_queues, _ = get_queues(app)
 
     # If not specified, get all the queues
     if queues is None:
         queues = [*active_queues]
+    # Celery adds the queue tag in front of each queue so we add that here
```
lucpeterson, line 493: Ah so here was the key bug 🐛

merlin/study/celeryadapter.py
```diff
@@ -424,23 +508,31 @@ def stop_celery_workers(queues=None, spec_worker_names=None, worker_regex=None):
...
+                    f"""Result of regex_list_filter for {worker_name}:
+                    {regex_list_filter(worker_name, all_workers, match=False)}"""
+                )
                 workers_to_stop += regex_list_filter(worker_name, all_workers, match=False)
+        # --workers flag
         if worker_regex is not None:
-            workers_to_stop += regex_list_filter(worker_regex, workers_to_stop)
+            for worker in worker_regex:
+                LOG.debug(f"Result of regex_list_filter: {regex_list_filter(worker, all_workers, match=False)}")
+                workers_to_stop += regex_list_filter(worker, all_workers, match=False)
```
lucpeterson, line 529: And here was the other one 🐞

lucpeterson (changes requested):
A few minor mods, but this looks great!

merlin/router.py
bgunnar5: I originally added this because pylint was complaining that there should be a return from every path in the function. I chose -1 for now as the return code to represent that something went wrong. This function in particular only returns a value for logging purposes so we could change this to not return anything and just log inside router.py rather than main.py.

Right now the return values from every function in router.py are not consistent (i.e. some return ints, some return lists, some don't return anything, etc.). The enum of return codes isn't a bad idea but it would require a slightly larger to-do as we'd have to refactor the return values of a lot of these functions. I'm all for creating a to-do for this, just warning that it might take more work than we think.

merlin/spec/specification.py
bgunnar5: I just went through and added comments so you'll see them after I push changes. Essentially, this just goes through and makes sure every step in a workflow is assigned to a worker. I found a bug where if I didn't assign every step to a worker then the step that I didn't assign would never get completed. For example, say I had 2 steps `step_1` and `step_2` and 1 worker `demo_worker`. In the `steps` key for the worker I assigned it `steps: [step_1]`. This would run my workflow but would never run `step_2` since there was never a worker assigned to it. With the above change, `step_2` would be assigned to the default worker and would then get completed even without specifically assigning it to a worker.

merlin/study/celeryadapter.py
bgunnar5: 🐛 🐞 🐛 🐞 🐛 🐞

merlin/spec/specification.py
lucpeterson: Ok! I see a little better now. Maybe I can suggest some variable name changes to help with confusion

merlin/spec/specification.py
```diff
@@ -354,8 +374,23 @@ def process_spec_defaults(self):
         if self.merlin["resources"]["workers"] is None:
             self.merlin["resources"]["workers"] = {"default_worker": defaults.WORKER}
         else:
-            for worker, vals in self.merlin["resources"]["workers"].items():
+            # Gather a list of step names defined in the study
+            steps = self.get_study_step_names()
```
lucpeterson, line 378: steps -> steps_workflow

merlin/spec/specification.py
```diff
@@ -354,8 +374,23 @@ def process_spec_defaults(self):
...
+            modified_steps = False
+            # Loop through each worker and fill in the defaults
+            for _, vals in self.merlin["resources"]["workers"].items():
```
lucpeterson, line 381: vals -> worker_settings

merlin/spec/specification.py
```diff
@@ -354,8 +374,23 @@ def process_spec_defaults(self):
...
                 MerlinSpec.fill_missing_defaults(vals, defaults.WORKER)
+                # Loop through the list of steps assigned to a worker
+                for step in vals["steps"]:
```
lucpeterson, line 384: step -> worker_step

merlin/spec/specification.py
```diff
@@ -354,8 +374,23 @@ def process_spec_defaults(self):
         if self.merlin["resources"]["workers"] is None:
             self.merlin["resources"]["workers"] = {"default_worker": defaults.WORKER}
         else:
-            for worker, vals in self.merlin["resources"]["workers"].items():
+            # Gather a list of step names defined in the study
+            steps = self.get_study_step_names()
+            modified_steps = False
```
lucpeterson, line 379: modified_steps -> assigned_steps_to_workers

merlin/spec/specification.py
```diff
@@ -354,8 +374,23 @@ def process_spec_defaults(self):
...
                 MerlinSpec.fill_missing_defaults(vals, defaults.WORKER)
+                # Loop through the list of steps assigned to a worker
+                for step in vals["steps"]:
+                    # If the assigned step is still in the list of existing steps, remove it
+                    if step in steps:
+                        modified_steps = True
+                        steps.remove(step)
+            # If other steps have been assigned a worker but there are still steps remaining
+            # that haven't been assigned a worker yet, assign the remaining steps to the default worker
+            if modified_steps and steps:
```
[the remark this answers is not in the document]
lucpeterson: You could also get fancy and condense a lot of this looping into the the subtraction of various sets. 

steps_that_need_workers = set(workfow_steps) - set(worker_steps)

If you want to be fancy 🎩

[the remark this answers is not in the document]
bgunnar5: I just did this, ran the test suite, and now I remember why I had the boolean there. 

The current logic is that if there are no workers defined in a spec then by default the `default_worker` is created and it has `steps: ["all"]` in it's worker settings. With the new code I'm adding, instead of it being `steps: ["all"]` it would be `steps: [<list of every step in workflow>]` which causes a bunch of tests to fail. 

Rather than having a bool to keep track of, I think I could just change the if statement to be something like `if steps_that_need_workers and (steps_that_need_workers != all_workflow_steps)`, where `all_workflow_steps` is a list of all the steps in the workflow (essentially `self.get_study_step_names()`).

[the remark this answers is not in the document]
bgunnar5: I just now saw your comment on using sets, I think that would be cleaner than a lot of this so maybe ignore the above message for now while I check that out.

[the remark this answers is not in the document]
bgunnar5: Using sets is a lot cleaner but I still needed that if check that I was talking about. I just pushed the new changes if you want to see them.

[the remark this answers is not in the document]
lucpeterson: Nice

lucpeterson (approved):

koning (approved):

PR #408, opened 2023-03-23
Fix pylint errors

bgunnar5:
This branch fixes almost all of the pylint errors we were seeing from the merlin/ and tests/ directories. I had to update the setup.cfg file to include some global disable flags for pylint.


lucpeterson (approved):
My only question is what was the logic for ignoring vs fixing errors? Perhaps make a note in the changelog, because many of them are “ignored” not “fixed” per se. A note in the changelog could help document what tech debt has been addressed / left for later

koning (approved):

bgunnar5:
I went back through this PR and cleaned up some additional pylint errors since I wasn't very consistent on when I fixed problems vs disabled them in my first run through. I added comments on some of the pylint errors I disabled and also added more info about pylint to the CHANGELOG.

The biggest changes you'll see are in `batch.py` but I think it makes the code easier to follow and easier to update going forward.

koning:
I think the batch.py refactor should be in a separate PR. This PR should be for pylint changes.

bgunnar5:
@lucpeterson it's test 14 in the local test suite that's failing here for some reason. It runs just fine locally for me so I'm trying to debug at the moment. Maybe I'll just change this PR to a draft until it's fully ready.

bgunnar5:
Ok I just removed the batch refactor from this branch so as long as these tests pass (they should) then this branch should be ready to merge.

koning (approved):

PR #409, opened 2023-03-23
Refactor query-workers

bgunnar5:
This branch was created based on a suggestion by Dylan Cliche. It adds 3 flags to `query-workers` that follow similar logic to `stop-workers`:
1. `--queues`: enables users to provide certain queues to query workers from
2. `--spec`: enables users to query workers from certain spec files
3. `--workers`: enables users to query workers by worker name (or regex)

Additionally, the output of this command has been changed from logging the worker names that are found to printing a table of workers and their associated queues.

### DIFF
--- a/merlin/study/celeryadapter.py
+++ b/merlin/study/celeryadapter.py
@@ -39,7 +39,7 @@ import time
 from contextlib import suppress
 
 from merlin.study.batch import batch_check_parallel, batch_worker_launch
-from merlin.utils import check_machines, get_procs, get_yaml_var, is_running, regex_list_filter
+from merlin.utils import apply_list_of_regex, check_machines, get_procs, get_yaml_var, is_running
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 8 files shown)


merlin/study/celeryadapter.py
```diff
@@ -407,9 +524,8 @@ def verify_args(spec, worker_args, worker_name, overlap, disable_logs=False):
         nhash = ""
         if overlap:
             nhash = time.strftime("%Y%m%d-%H%M%S")
-        # TODO: Once flux fixes their bug, change this back to %h
         # %h in Celery is short for hostname including domain name
-        worker_args += f" -n {worker_name}{nhash}.%%h"
+        worker_args += f" -n {worker_name}{nhash}.%h"
```
lucpeterson, line 528: Don’t change this until we’ve confirmed it works

merlin/study/celeryadapter.py
```diff
@@ -488,7 +604,6 @@ def stop_celery_workers(queues=None, spec_worker_names=None, worker_regex=None):
 
     """
     from merlin.celery import app  # pylint: disable=C0415
-    from merlin.config.configfile import CONFIG  # pylint: disable=C0415
```
lucpeterson, line 491: I think moving this early is what broke the tests

merlin/study/celeryadapter.py
bgunnar5: I tried it out with the flux_par example and it seemed to work. I can switch it back for now though.

merlin/study/celeryadapter.py
bgunnar5: I think you're right. I just encountered the same issue with the pylint-errors branch

bgunnar5:
For documentation purposes: @lucpeterson is on vacation atm but he approved this merge.

PR #411, opened 2023-03-29
refactor/batch

bgunnar5:
This branch refactors the batch.py file to separate functionality in a more organized fashion and condense similar code. These changes will make maintaining/updating this file easier in the future.


koning:
This code refactors the worker launch, so it can easily be tested using string comparison of the command with the --echo argument. The fake executables are already setup to mimic the various schedulers and have tests. Just need to add a fake jsrun for lsf to test all paths.

bgunnar5:
This should be good to merge now. It says the changelog isn't updated but I accidentally left the stuff related to the batch refactor in there when I merged the pylint fixes so that can be ignored.

bgunnar5:
@koning Ok I'll look into that

bgunnar5:
For documentation purposes I'm leaving this comment here: @lucpeterson and I discussed the issue with walltime that @koning mentioned above and determined it'd be best to put it on another branch so I'm merging this branch now (with Luc's permission).

Issue #412, opened 2023-03-30
Walltime in Batch Block


bgunnar5:
## Bug Report

**Describe the bug**
When you define a walltime in the batch block, if it's over 10 hours yaml converts it from HMS format to seconds. For example, `walltime: 10:00:00` is converted to `walltime: 36000`.

**To Reproduce**
1. Place `walltime: 10:00:00` or a larger walltime in the batch block of a spec file (does not matter the type this happens for flux, slurm, etc.)
2. Run `merlin run-workers <spec file> --echo`
3. Investigate the resulting output command and see the incorrect walltime

**Expected behavior**
Defining `walltime: 10:00:00` or greater should stay in the HMS format until it needs to be converted on a scheduler by scheduler basis.

**Additional context**
@lucpeterson sent me [this stack overflow solution](https://stackoverflow.com/a/37958106) which would be a good idea to try. We also need to add a test for this which might even be as simple as adding an extra condition to some existing tests (look at test ids 11-18 in the integration test suite I think).

bgunnar5:
I've been playing around with a fix for this issue by messing with PyYAML's implicit resolvers based on [the stack overflow solution](https://stackoverflow.com/a/37958106) from above. It turns out it's not the timestamp implicit resolver that's converting something like `10:00:00` to `36000`, instead it's the int implicit resolver. When I remove the int implicit resolver it fixes the walltime issue but causes issues elsewhere when loading the yaml for values that _are_ actually ints. So unless we want to go through and manually convert all of these values to ints (I would rather not do this) then this option doesn't seem reasonable.

I've also tried creating my own implicit resolvers for walltime but since the colon character is special in yaml files, it's seemingly impossible to create a regex that finds a match to the `HH:MM:SS` format. This option might work if we parse the yaml file first and convert the colons here to semi-colons (i.e. `HH;MM;SS`) and then use an implicit resolver on that, but not sure if that's the best idea either.

I did find a hacky solution that works by adding:
```
    walltime_match = re.search(r"(?:(?:([0-9][0-9]|2[0-3]):)?([0-5][0-9]):)?([0-5][0-9])", obj)
    if walltime_match:
        obj = f'"{obj}"'
```

To the end of the `_process_string` method in the MerlinSpec class.

Not sure what the best option is here.

bgunnar5:
Ok so disregarding messing with PyYAML I've now found two sort of hacky solutions.

1. Adding the following code to the end of `_process_string` in the `MerlinSpec` class:
[...]

PR #414, opened 2023-04-03
bugfix/walltime-conversion

bgunnar5:
This is still a draft but it's a start to a solution for #412. In the comments on that issue I came up with two solutions and I'm using the one I think is best. I'm open to feedback and willing to change this if we think there's a better solution.

This is most likely an issue that I'll revisit in the future since @jwhite242 and I have talked about creating a one-size-fits all walltime value in Maestro at some point.


koning:
The flux calls need a different time format, should have tests for slurm, flux and pbs for worker alloc as well as run:

flux:
      -t, --time-limit=FSD
              Set a time limit for the job in Flux standard duration (RFC 23).
              FSD is a floating point number with  a  single  character  units
              suffix  ("s", "m", "h", or "d"). If unspecified, the job is sub-
              ject to the system default time limit.

slurm:
      -t, --time=<time>
              Set a limit on the total run time of the job allocation.  If the
              requested time limit exceeds the partition's time limit, the job
              will  be  left  in a PENDING state (possibly indefinitely).  The
              default time limit is the partition's default time limit.   When
              the  time  limit  is reached, each task in each job step is sent
              SIGTERM followed by SIGKILL.  The interval  between  signals  is
              specified  by  the  Slurm configuration parameter KillWait.  The
              OverTimeLimit configuration parameter may permit the job to  run
              longer than scheduled.  Time resolution is one minute and second
              values are rounded up to the next minute.

              A time limit of zero requests that no  time  limit  be  imposed.
              Acceptable  time  formats  include "minutes", "minutes:seconds",
              "hours:minutes:seconds", "days-hours", "days-hours:minutes"  and
              "days-hours:minutes:seconds".

pbs:
    qsub -l walltime=<hours:minutes:seconds> script

koning:
Yes, that function is used in the scriptadapter but not the batch alloc yet.

bgunnar5:
Whoops I somehow didn't see that the `convert_timestring` function took in ints, I'll try switching to that.

Should we just ignore the walltime value if lsf is the scheduler since it doesn't seem to have a walltime flag? I did a little digging and they do have a walltime flag (-W) for bmod but not one for jsrun that I can find.

koning:
I think your original plan to keep the string in the resulting yaml is a good one. We can just use convert_timestring internally. I don't think jsrun has a walltime flag either. On LSF you really need the bsub/jsrun combo. So the ultimate solution will be to emit a bsub script with the jsrun launch.

bgunnar5:
I have a feeling that's going to require a lot of changes so maybe we hold off on that in this PR?

koning:
Yes, that is a future improvement and a low priority.

PR #425, opened 2023-05-24
Bugfix/filename-special-vars

### LINKED ISSUE #423: [BUG] MERLIN_SPEC variables are not working properly
## Bug Report

**Describe the bug**
`$(MERLIN_SPEC_ORIGINAL_TEMPLATE)`, `$(MERLIN_SPEC_EXECUTED_RUN)`, and `$(MERLIN_SPEC_ARCHIVED_COPY)` do not rename the specification's filename with extension `.orig.yaml`, `.partial.yaml`, or `.expanded.yaml` respectively. Instead, it is using the `name:` given in the `description:` block within the specification file.

**To Reproduce**
In any workflow change the name in the description block to not be the filename of the specification file and type `echo <any of the Merlin variables given above>` in any of the workflow's steps.

**Expected behavior**
In the output file for the step that contained the echo command, it should contain `<spec filename>.<orig or partial or expanded>.yaml`.

**Please answer these questions to help us pinpoint the problem**
- Does the problem occur in `merlin run --local` mode, distributed or neither? **Both**
- On what machines/architectures are you running merlin? Is this bug on a specific machine or can you reproduce it elsewhere? **Haven't tried**

**Please run `merlin info` and paste the results here:**

Merlin Configuration
-------------------------

 config_file        | /g/g15/cliche1/.merlin/app.yaml
 is_debug           | False
 merlin_home        | /g/g15/cliche1/.merlin
 merlin_home_exists | True
 broker server      | amqps://cliche1:******@rzrabbit.llnl.gov:5671/cliche1
 broker ssl         | True
 results server     | redis://mlsi:******@rzrabbit.llnl.gov:6379/0
 results ssl        | False

Checking server connections:
----------------------------
broker server connection: OK
results server connection: OK

bgunnar5:
This branch fixes #423 in which the .orig, .partial, and .expanded yaml copies in the merlin-info subdirectory were being named after the study rather than after the name of the spec file. This branch also adds unit tests for this issue and modifies 3 integration tests to accommodate this fix.


koning (approved):

merlin/study/study.py
```diff
@@ -413,23 +415,21 @@ def expanded_spec(self):
...
         # write original spec for provenance
-        result = MerlinSpec.load_spec_from_string(result.dump())
-        result.path = expanded_filepath
-        name = result.description["name"].replace(" ", "_")
-        self.write_original_spec(name)
+        self.write_original_spec()
 
         # write partially-expanded spec for provenance
         partial_spec = deepcopy(self.original_spec)
+        result = MerlinSpec.load_spec_from_string(result.dump())
```
lucpeterson, line 426: what are these lines doing?

lucpeterson (approved):
minor comments; fix and go!

merlin/study/study.py
bgunnar5: I didn't write these lines myself, I just copied them from the existing code and moved them down to organize the code better. It looks like before this line, result is building the entire expanded spec and then creating a MerlinSpec object from all the info loaded in. I'm not sure why it's necessary to create an entirely new MerlinSpec object here if result is already a MerlinSpec object... I'll try removing this and see if it impacts the partial spec creation at all

bgunnar5: After commenting them out, the partial yaml file is still outputting the same information so I don't think these lines are necessary. I'm removing them

bgunnar5: oop I lied, setting the path of the result is necessary for testing so I'm putting that line back in there

PR #426, opened 2023-06-15
bugfix/openfoam_singularity_issues

bgunnar5:
Fixed a bug where the openfoam singularity example was unable to be found due to an incorrect filename. I also updated the scikit requirement version in the requirements.txt of this example to eliminate a setup error.


koning:
I think the requirements in the other openfoam examples need to be updated as well.

merlin/examples/workflows/openfoam_wf_singularity/requirements.txt
```diff
@@ -1,3 +1,3 @@
 Ofpp==0.11
-scikit-learn==0.21.3
```
koning, line 2: This can just be >=0.21.3 to capture the old behavior.

bgunnar5:
I'm seeing some strange behavior at the moment: 
I tried running the openfoam_wf example with the updated requirement for scikit-learn set to >=0.21.3. This caused the setup step to crash again but then the sim runs step proceeded to execute. The sim runs step stalled once each worker pulled a sample to execute so I then stopped the workers and purged the queues. After that, I tried another run with the scikit-learn set to >=1.0.2 to see if that fixed the issue, but when I ran my workers again they started executing those same stalled steps from the last workflow run and ended up stalling again.

I'm not sure why this is happening considering I stopped the workers _and_ purged the queues... I would expect those stalled tasks to be purged and not in the celery queue to be executed again.

merlin/examples/workflows/openfoam_wf_singularity/requirements.txt
koning: Does this need to be changed in the other openfoam examples as well?

bgunnar5: I think so. I just tried this again and it seems to work fine now. Not sure why/how I was seeing the strange behavior I mentioned below... I'll get the requirements files changed so we can get this merged.

koning:
The openfoam_wf is the one with docker and the one in openfoam_wf_no_docker is for a local openfoam install. I see files on that dir in the repo.

For the task issue, two possibilities to check are that the workers were not fully stopped before you purged, or you may have (non-executing ) workers pulling from the same queues (default queue) elsewhere on the systems.

merlin/examples/workflows/openfoam_wf_singularity/requirements.txt
bgunnar5: > This can just be >=0.21.3 to capture the old behavior.

@koning I've been thinking about this more and since the setup error was happening in v0.21.3 I'm not sure if we should keep the = sign here in the requirements file. In other words I think the version has to be newer than 0.21.3 in order to avoid the setup error we were seeing

koning: OK, just trying to keep the other installs consistent. If the setup is the issue, then it should be updated.

koning (approved):

PR #427, opened 2023-06-27
feature/pdf-docs

bgunnar5:
Added auto pdf creations on each docs build. Users can download this pdf through the dropdown menu in the docs.


bgunnar5:
@doutriaux1 Do you mean adding a similar feature to the weave_ci docs or is there already a pdf for those that you want me to link to?

doutriaux1:
> @doutriaux1 Do you mean adding a similar feature to the weave_ci docs or is there already a pdf for those that you want me to link to?

I meant in the weave_ci docs where we point to the merlin docs, we should add a link to the generated pdf.

But now that you mention it it might be a good idea to add this to the whole weave_ci as well. Not sure what the pdf would look like though.

koning (approved):

.readthedocs.yaml
```diff
@@ -11,3 +11,5 @@ sphinx:
 python:
   install:
     - requirements: docs/requirements.txt
+
+formats: [pdf]
```
koning, line 15: Do you need html here also or will readthedocs build the html target?

bgunnar5: Readthedocs will build the html target by default, this is just for additional formats you wish to build. You can see the other formats [here](https://docs.readthedocs.io/en/stable/config-file/v2.html#formats).

PR #430, opened 2023-07-10
bugfix/output-path-substitution

bgunnar5:
A user named Shailaja brought a bug to my attention that involved CLI substitutions for variables used in the `OUTPUT_PATH` variable. For example, say you have an environment block like so:
```
env:
    variables:
        DATE: temp
        OUTPUT_PATH: study_$(DATE)
```
If you run this study with no CLI substitutions, the output path is set as intended (i.e. it's named study_temp); however, if you try to set the `DATE` variable from the CLI with `merlin run <spec file> --vars DATE="date +%Y%m%d-%H%M%S"`, the `OUTPUT_PATH` variable is still named study_temp which is not expected behavior.

This PR fixes that issue.


koning:
is there a test this can be added to?

bgunnar5:
@koning good call. I'll check and add one

koning:
OK, otherwise , looks good to me.

koning (approved):

PR #431, opened 2023-07-12
refactor/status

bgunnar5:
This is the branch with all of the status command changes that I've been working on this year. I'm making this a draft PR for now so that we can get the review process started since it's a huge PR. I'm still working on finishing the documentation, and I still need to add tests.

This branch:
- Introduces a new `merlin queue-info` command
  - this contains the same behavior as the previous status command
  - some additional behavior as well (active queue viewing, specific queue viewing, etc.)
- Refactors the current `merlin status` command
  - now shows a step-by-step summary of the status of a study
  - includes a progress bar and a summary block for each step
  - additional options to remove prompts and add colorblind assistance
- Introduces a new `merlin detailed-status` command
  - shows task-by-task statuses of a study
  - allows for filtering by return codes, task queues, task statuses, and workers
  - can set a limit on how many tasks to display
  - integrates Maestro's status display format and provides options to modify this display

There are still a couple of tickets for this that I have yet to implement but I figured they could be done after we get this released so users can start trying it out. Those tickets include:
- Replacing MERLIN_FINISHED files with reads from MERLIN_STATUS.json files
- Creating new status entries for each restart/retry
  - currently we're just counting the number of restarts which isn't super useful
  - I recently made the change from .txt files to .json files to help make this feature easier to implement
- Condensing detailed-status output a bit more (I'm still planning out how to do this)
  - right now the output will show every single task and the info associated
  - maybe we can put tasks with similar statuses together?
    - e.g. all tasks with MERLIN_SUCCESS return code can be grouped together
  - maybe we don't need to show elapsed/run times and number of restarts for tasks grouped together?


bgunnar5:
I may end up splitting this PR into a couple different PRs (maybe one for each new command) to make it easier to review. If you'd prefer that then let me know and I'll work on it

koning:
The one item missing from `merlin monitor` is a poll for remaining in flight tasks . That is, the queue is empty but the workers are currently running some tasks. Is it possible to have that functionality exposed for the monitor?

koning:
That's fine, I just wanted to make you aware in case some current code can be exposed for that purpose.

merlin/utils.py
```diff
@@ -497,6 +498,31 @@ def contains_shell_ref(string):
     return False
 
 
+def dict_deep_merge(dict_a, dict_b, path=None):
+    """
+    This function recursively merges dict_b into dict_a. The built-in
```
jwhite242, line 503: Would be useful to make it clear in the docstring that this only merges in new keys and will not update existing one (is there a need for an 'update' version of this that also updates existing keys?)

merlin/utils.py
```diff
@@ -497,6 +498,31 @@ def contains_shell_ref(string):
...
+    if path is None:
+        path = []
+    for key in dict_b:
+        if key in dict_a:
+            if isinstance(dict_a[key], dict) and isinstance(dict_b[key], dict):
+                dict_deep_merge(dict_a[key], dict_b[key], path + [str(key)])
+            elif dict_a[key] == dict_b[key]:
+                pass  # same leaf value
+            else:
+                raise Exception(f"Conflict at {'.'.join(path + [str(key)])}")
```
jwhite242, line 521: A more custom exception type should be used here I think.  Raising base exception type makes it hard to catch and handle just this one without also catching any other unrelated exception thrown in this scope.  Could be as simple as:

```
class DeepMergeException(Exception):
    pass
```
Or you can extend it to take more args and do more custom messaging (may need to implement `__reduce__` method too to ensure it can be pickled if you start adding more args).

merlin/utils.py
```diff
@@ -560,3 +586,19 @@ def convert_timestring(timestring: Union[str, int], format_method: str = "HMS")
     tdelta = convert_to_timedelta(timestring)
     LOG.debug(f"Timedelta object is: {tdelta}")
     return repr_timedelta(tdelta, method=format_method)
+
+
+def ws_time_to_td(ws_time: str) -> datetime:
```
jwhite242, line 591: should it be 'dt' instead of 'td', as in short for 'datetime'?

tests/integration/conditions.py
```diff
@@ -229,7 +229,7 @@ def contains(self):
             with open(filename, "r") as textfile:
                 filetext = textfile.read()
             return self.is_within(filetext)
-        except Exception:  # pylint: disable=W0718
+        except Exception:  # pylint: disable=W0703
```
jwhite242, line 232: Are there more specific exceptions that could/should be caught here like filenotfound io/etc?  I.e. what's this exception actually meant to handle and why is False always the right return value?

merlin/utils.py
bgunnar5: I'll add that to the docstring here, thanks for pointing it out. In all the cases that I use this function I don't think I ever need to update existing keys. It's mainly used to help store all statuses that need to be read in/modified in one dict.

merlin/utils.py
bgunnar5: yes good catch

tests/integration/conditions.py
bgunnar5: I plan to create a branch in the future to eliminate as many of these "pylint: disable" messages as possible (there's a decent amount throughout our codebase) so for now I'm going to leave this as is, but I'll come back to it and clean it up in the future.

jwhite242:
More of a general comment that's not for a specific file: can you generate the test study without the timestamp?  would make for less churn as the test suite evolves and things need updating (maybe even a different convention based on merlin version numbering to enable some backward compatibility testing)

bgunnar5:
I think it'd take some modification in the workspace method of the MerlinStudy class but it could be done

bgunnar5:
@lucpeterson @koning @doutriaux1 @jwhite242 I'd like your opinions here. I wrote a new unit test file called `test_celeryadapter.py` in order to test the functions that exist in the `celeryadapter.py` module (specifically the ones related to the `merlin queue-info` command for this branch). A lot of the functions in this module require an existing Celery app and, by association, an active Redis server. Therefore, writing tests for these functions required me to spin up a dummy Celery app and local Redis server. Since these tests depend on outside systems (Celery/Redis), would it be best to move them over to the integration test suite even though they're testing singular functions?

The GitHub action is failing at the moment since I'm starting a Redis server using the `redis-server` command, which it doesn't have access to. If I move these tests over to the integration test suite this would no longer be required, but I'm not sure if we should be testing singular functions within the integration test suite or not. What do you guys think?

koning:
Maybe these tests can be rolled into the merlin server tests, since that is starting a server?

bgunnar5:
That's not a bad idea. I'll try spinning the server up in the base class with the `merlin server` command instead of the `redis-server` one which should resolve the issue with the github action. Thanks for the suggestion

PR #442, opened 2023-08-18
feature/new-status

bgunnar5:
I decided it would be best to split the refactor/status branch into 4 different branches:

1. A branch for the `merlin status` command
2. A branch for the `merlin detailed-status` command
3. A branch for the `merlin queue-info` command
4. A branch for all of the documentation for these commands

This is the branch for the `merlin status` changes and will most likely be the largest of the 4.

The refactored `merlin status` command in this PR contains the following functionality:
- now shows a step-by-step summary of the status of a study
- includes a progress bar and a summary block for each step
- additional options to remove prompts and add colorblind assistance


tests/unit/study/status_test_files/status_test_variables.py
```diff
@@ -0,0 +1,471 @@
...
+# SOFTWARE.
+###############################################################################
+"""This module holds variables that will be used to test against output from calls to status methods"""
+import os
+
+
+# Global path variables for files we'll need during these status tests
+PATH_TO_TEST_FILES = f"{os.path.dirname(__file__)}"
+SPEC_PATH = f"{PATH_TO_TEST_FILES}/status_test_spec.yaml"
+VALID_WORKSPACE = "status_test_study_20230717-162921"
```
lucpeterson, line 37: Does this mean all these empty files are here for a reason and not just a typo? If so, can we rename the directories so they look like they’re supposed to be there. The date stuff is throwing me off, as I first thought this was a simulation run

lucpeterson:
Overall this looks fairly complete, with new and associated tests passing. The style seems good, with sufficient modularity. I think I’m ok with this (except above which is a nice to have): any detailed investigation would be of performance, scalability, etc and can be probably explored with additional improvements. In terms of a minimum viable product, I think this will suffice!

lucpeterson (changes requested):
The big change I’d like to see is any mods to the documentation (unless those are in a separate PR)

bgunnar5:
@lucpeterson I'll throw the docs in my final PR for this. If you want a look at them you can see them at #431

tests/unit/study/status_test_files/status_test_variables.py
bgunnar5: @lucpeterson Ok I remember why I named the directories with time stamps rather than giving them "test" names: the `merlin status` command will build a list of potential studies for the user to view the status of and part of how it does this is to use a regex for the timestamp that gets appended. In order to test that this functionality works correctly we need the timestamp in the names of these directories.

koning:
I tried this PR, looks good.

lucpeterson (approved):

koning (approved):

PR #447, opened 2023-09-22
feature/vlauncher

bgunnar5:
Added a new feature that was requested by @koning for the ICECap team. This feature is the `$(VLAUNCHER)` command. Instead of reading allocation configuration from the step.run block, it instead reads specific shell variables. The variables that it reads are:
- `MERLIN_NODES`
- `MERLIN_PROCS`
- `MERLIN_CORES`
- `MERLIN_GPUS`

Currently, the GPU option only works with flux. However, I'll be modifying the script adapters soon and this will be fixed.

Also, I fixed a minor naming error that happened on iterative workflows. This bug caused filenames for the .out, .partial, and .expanded files to continuously append .out, .partial, or .expanded (depending on which was used in the iteration step) to the name of the file.


koning:
@bgunnar5 Thanks this is a good start. I added a new branch, PR_447, that will use the variable names instead of the variable values. The values may not be set until the script is run, so they cannot be replaced when the task is defined.

merlin/spec/defaults.py
```diff
@@ -52,3 +52,10 @@
     "generate": {"cmd": "echo 'Insert sample-generating command here'"},
     "level_max_dirs": 25,
 }
+
+VLAUNCHER_VARS = {
+    "MERLIN_NODES": 1,
```
koning, line 57: Are these overridden by the step values?  WE should let the step values have precedence.

bgunnar5:
> @bgunnar5 Thanks this is a good start. I added a new branch, PR_447, that will use the variable names instead of the variable values. The values may not be set until the script is run, so they cannot be replaced when the task is defined.

This is a good point. When I did this last Thursday I just set it up to use the most recently defined MERLIN_* variables but that might not always be the case. Did you have a chance to test out your changes on the PR_447 branch and do they accomplish what you need?

koning:
Yes, the PR_447 branch will replace the names and not values and also check for csh to set the variables.

koning:
Would you please merge the PR_447 branch into this PR and add the step defaults as the MERLIN_<VAR> defaults?

bgunnar5:
@koning yes I will. I think it will also be a good idea to add something to check if the line with MERLIN_* defined is commented out and if it is then ignore it

PR #450, opened 2023-10-09
bugfix/skewed-sample-hierarchy

bgunnar5:
This branch has the patch for the skewed sample hierarchy bug. This bug would happen when a restart was needed in the `add_merlin_expanded_chain_to_chord` in the recursive section. If an exception was raised that triggered a restart here, then the path where recursion is started would not be reset correctly. This leads to skewed sample hierarchy paths. For example, 06/00/01 could become 06/06/00/01.

This bug also lead to extra samples being ran since these skewed hierarchies could create many new paths.

### DIFF
--- a/merlin/common/tasks.py
+++ b/merlin/common/tasks.py
@@ -321,7 +321,7 @@ def add_merlin_expanded_chain_to_chord(  # pylint: disable=R0913,R0914
                 else:
                     self.add_to_chord(next_step, lazy=False)
                 LOG.debug(f"queued for samples[{next_index.min}:{next_index.max}] in for {chain_} in {next_index.name}")
-        except Exception as e:
+        except retry_exceptions as e:
             # Reset the index to what it was before so we don't accidentally create a bunch of extra samples upon restart
             next_index.name = next_index_name_before
             raise e
--- a/merlin/common/tasks.py
+++ b/merlin/common/tasks.py
@@ -298,27 +298,33 @@ def add_merlin_expanded_chain_to_chord(  # pylint: disable=R0913,R0914
         LOG.debug("chain added to chord")
     else:
         # recurse down the sample_index hierarchy
-        LOG.debug("recursing down sample_index hierarchy")
-        for next_index in sample_index.children.values():
-            next_index.name = os.path.join(sample_index.name, next_index.name)
-            LOG.debug("generating next step")
-            next_step = add_merlin_expanded_chain_to_chord.s(
-                task_type,
-                chain_,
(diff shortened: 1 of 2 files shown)


merlin/common/tasks.py
```diff
@@ -298,27 +298,33 @@ def add_merlin_expanded_chain_to_chord(  # pylint: disable=R0913,R0914
...
+                )
+                next_step.set(queue=chain_[0].get_task_queue())
+                LOG.debug(f"recursing with range {next_index.min}:{next_index.max}, {next_index.name} {signature(next_step)}")
+                LOG.debug(f"queuing samples[{next_index.min}:{next_index.max}] in for {chain_} in {next_index.name}...")
+                if self.request.is_eager:
+                    next_step.delay()
+                else:
+                    self.add_to_chord(next_step, lazy=False)
+                LOG.debug(f"queued for samples[{next_index.min}:{next_index.max}] in for {chain_} in {next_index.name}")
+        except Exception as e:
```
doutriaux1, line 324: Can we predict the exception type? Or do we have to stay with `Exception`?

doutriaux1 (changes requested):
@bgunnar5 thanks for catching this! I'm a stickler for always adding a test before approving a PR. Also I like to see the docs updated (but in this case I don't think it applies).

merlin/common/tasks.py
bgunnar5: There are a wide range of exceptions that could happen here so I believe that leaving this as broad as possible is for the best.

bgunnar5:
@doutriaux1 This one is tricky to add a test for as it's in the middle of a celery task. I could maybe add a kwarg specifically for testing to this task (e.g. `test_mode=True`) that will automatically trigger an exception and therefore a restart.

I think this will get easier to test when I update the test suite to use pytest/fixtures.

merlin/common/tasks.py
bgunnar5: I might actually pull back on what I just said here. We have a tuple of exceptions defined that will all trigger task restarts. I'm going to see if it's possible to just provide the variable with all of the exceptions rather than using `Exceptions`

doutriaux1:
> @doutriaux1 This one is tricky to add a test for as it's in the middle of a celery task. I could maybe add a kwarg specifically for testing to this task (e.g. `test_mode=True`) that will automatically trigger an exception and therefore a restart.
> 
> I think this will get easier to test when I update the test suite to use pytest/fixtures.

@bgunnar5 if you don't add a test now please make a note to add one to the PR where you will switch to pytest.

PR #451, opened 2023-10-10
feature/detailed-status

bgunnar5:
This branch contains the functionality and tests for the new `merlin detailed-status` command. This command displays more detailed status information for each step. This command will allow users to filter by status, return code, worker, task queue, and steps. Users can also control how many tasks are displayed in total to help keep the output to a minimum.

The display functionality adopts from Maestro's status renderers, which allowed me to create two different display layouts for this command.

The docs for this will be released later but you can get a sneak peak at #431

### DIFF
--- a/merlin/main.py
+++ b/merlin/main.py
@@ -55,7 +55,8 @@ from merlin.log_formatter import setup_logging
 from merlin.server.server_commands import config_server, init_server, restart_server, start_server, status_server, stop_server
 from merlin.spec.expansion import RESERVED, get_spec_with_expansion
 from merlin.spec.specification import MerlinSpec
-from merlin.study.status import Status
+from merlin.study.status import VALID_RETURN_CODES, VALID_STATUS_FILTERS, DetailedStatus, Status
+from merlin.study.status_renderers import status_renderer_factory
 from merlin.study.study import MerlinStudy
 from merlin.utils import ARRAY_FILE_FORMATS, verify_dirpath, verify_filepath
 
@@ -191,11 +192,11 @@ def launch_workers(args):
     spec, filepath = get_merlin_spec_with_override(args)
     if not args.worker_echo_only:
         LOG.info(f"Launching workers from '{filepath}'")
-    status = router.launch_workers(spec, args.worker_steps, args.worker_args, args.disable_logs, args.worker_echo_only)
+    launch_worker_status = router.launch_workers(spec, args.worker_steps, args.worker_args, args.disable_logs, args.worker_echo_only)
     if args.worker_echo_only:
-        print(status)
+        print(launch_worker_status)
     else:
-        LOG.debug(f"celery command: {status}")
+        LOG.debug(f"celery command: {launch_worker_status}")
 
 
 def purge_tasks(args):
@@ -255,11 +256,10 @@ def query_status(args):
         args.spec_provided = spec_provided
 
     # Get either a Status object or DetailedStatus object
-    # if args.detailed:
-    #     status_obj = DetailedStatus(args, spec_display, file_or_ws)
-    # else:
-    #     status_obj = Status(args, spec_display, file_or_ws)
-    status_obj = Status(args, spec_display, file_or_ws)  # The above can be uncommented when we add DetailedStatus
+    if args.detailed:
+        status_obj = DetailedStatus(args, spec_display, file_or_ws)
+    else:
+        status_obj = Status(args, spec_display, file_or_ws)
 
     # Handle output appropriately
     if args.dump:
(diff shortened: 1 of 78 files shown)


lucpeterson (commented):
Can you post a screenshot of the new feature?

bgunnar5:
When you run `merlin detailed-status <study>` it should pull up a pager window that looks like so:
detailed-status

For failures the entire row will be highlighted red to ensure that they stand out:
detailed-status-failed

bgunnar5:
@lucpeterson see above, I forgot to tag you in it

merlin/main.py
```diff
@@ -929,6 +932,99 @@ def generate_diagnostic_parsers(subparsers: ArgumentParser) -> None:
...
+    detailed_status: ArgumentParser = subparsers.add_parser(
+        "detailed-status",
+        help="Display a task-by-task status of a study.",
+    )
+    detailed_status.set_defaults(func=query_status, detailed=True)
+    detailed_status.add_argument(
+        "spec_or_workspace", type=str, help="Path to a Merlin YAML spec file or a launched Merlin study"
+    )
+    detailed_status.add_argument(
+        "--dump", type=str, help="Dump the status to a file. Provide the filename (must be .csv or .json).", default=None
```
jwhite242, line 945: Could default here be the same as the study name/timestamped variant, with some suffix maybe for the time said status gets dumped in case this gets called repeatedly?

merlin/main.py
```diff
@@ -929,6 +932,99 @@ def generate_diagnostic_parsers(subparsers: ArgumentParser) -> None:
...
+                            Default: %(default)s",
+    )
+    detailed_status.add_argument(
+        "--vars",
+        action="store",
+        dest="variables",
+        type=str,
+        nargs="+",
+        default=None,
+        help="Specify desired Merlin variable values to override those found in the specification. Space-delimited. "
```
jwhite242, line 961: What does it mean to override spec variables for the status command specifically?  Isn't that already captured in the study workspace and not to be changed further?

bgunnar5: Yes but with the status/detailed-status command I implemented you can pass in a spec file as well as an output directory (e.g. `merlin status my_study.yaml` or `merlin status my_study_20231010-140705/`). When a spec file is passed in, Merlin will do some behind the scenes work and look for any potential studies that exist w.r.t. that spec file. It will then ask the user which one they want to view the status of:
status-potential-studies

It does all of this by looking in the `OUTPUT_PATH` for directories that match the name of the study and have a time stamp. So, if a user modified the `OUTPUT_PATH` variable at the CLI for their run then they would need to modify the `OUTPUT_PATH` again if they decided to pass a spec in as input to the status command or else Merlin would be unable to find the study they're looking for. If they pass in an output directory rather than a spec file then this no longer matters, but I was just trying to think of any possible scenario users might run into with this format. Does this make sense?

merlin/main.py
bgunnar5: I'm not sure if I would have access to the name of the study at this point in the code since it's at the very top level of our codebase. In other words, I'm not sure if I'd be able to access the yaml file/output directory that the user passes in since at this point the arguments will not have been parsed yet. 

My thinking was that this would just default to `None` to represent that no dumping will take place unless the user specifically decides to dump the status.

merlin/main.py
bgunnar5: Thinking this through a bit more, I'm pretty sure `OUTPUT_PATH` would be the only variable that would need to be manipulated at the CLI for `status` and `detailed-status`. To make this more clear, maybe I should just add a `--output-path` option and remove the `--vars` option completely?

bgunnar5:
@lucpeterson @koning Do you think this branch is good to merge? It says the CHANGELOG is not updated but it is

merlin/router.py
```diff
@@ -236,43 +240,130 @@ def create_config(task_server: str, config_dir: str, broker: str, test: str) ->
...
+
+    # Get the active queues and the workers that are watching them
+    active_queues = get_active_queues(args.task_server)
+    LOG.debug(f"Monitor: active_queues: {active_queues}")
+
+    # Count the number of workers that are active
+    consumers = set()
+    for active_queue, workers_on_queue in active_queues.items():
+        if active_queue in queues_in_spec:
+            consumers |= set(workers_on_queue)
```
lucpeterson, line 351: Can you explain the bit wise OR operation here?

bgunnar5: The `|` operator in python can be used for taking the union of two sets. Here, I'm doing an in-place union of the `consumers` set and the set of `workers_on_queue`. For example, if I have:
```
consumers = {"worker1", "worker2", "worker3"}
workers_on_queue = {"worker3", "worker5"}
```
Then, doing the in-place union with `consumers |= workers_on_queue` would change `consumers` value to `{"worker1", "worker2", "worker3", "worker5"}`

lucpeterson: Gotcha! I hadn’t seen that use case for set union

lucpeterson (approved):

koning (approved):

PR #453, opened 2023-10-19
bugfix/lsf-gpu-typo

bgunnar5:
There was a typo in `batch.py` that resulted in worker launch failures on blueos machines for workflows that try to use flux. This branch fixes that typo.

### DIFF
--- a/merlin/study/batch.py
+++ b/merlin/study/batch.py
@@ -299,7 +299,7 @@ def construct_scheduler_legend(parsed_batch: Dict, nodes: int) -> Dict:
         "lsf": {
             "check cmd": ["jsrun", "--help"],
             "expected check output": b"jsrun",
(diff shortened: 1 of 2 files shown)


merlin/study/batch.py
```diff
@@ -335,12 +335,16 @@ def construct_worker_launch_command(parsed_batch: Dict, nodes: int) -> str:
     scheduler_legend: Dict = construct_scheduler_legend(parsed_batch, nodes)
     workload_manager: str = get_batch_type(scheduler_legend)
 
+    print(f"parsed_batch: {parsed_batch}")
```
koning, line 338: Do you want to leave in these print statements? Maybe they should have logger if clauses?

bgunnar5: whoops I didn't mean to leave this in there, good catch

koning (approved):

PR #457, opened 2023-11-07
Bugfix for WEAVE CI

bgunnar5:
The WEAVE CI was running into issues running the new tests that use the pytest fixtures I introduced last week. This is due to the issue with the `merlin server` command where you have to run `export LC_ALL="C"` before starting the server. This branch puts that export command directly in the test suite so we no longer have to worry about it causing issues.


lucpeterson:
Is this a real fix or just a band-aid on a fundamental problem with “merlin server”?

doutriaux1 (approved):
Except for the chagelog complaint thais is one should be good to go

bgunnar5:
@lucpeterson more of a band-aid. This is specifically to get the tests to pass on WEAVE's CI since I don't have time to look into what's really causing this for the `merlin server` command at the moment

PR #459, opened 2024-01-19
Add the missing restart keyword to the specification docs.

koning:


bgunnar5 (approved):
@koning This looks fine to merge.

I will say I'm working on a large documentation overhaul that's almost finished and I've added my own little blip about the `restart` key:

image

However, I still think this should be merged as I'm not sure how long the new documentation will take to review.

koning:
OK, I wanted to make sure it was there for reference.

PR #463, opened 2024-02-12
openfoam tutorial modifications

xorJane:
Greetings! This PR changes the Merlin tutorial's section, "Run a Real Simulation", to default to using an OpenFoam container with singularity rather than docker. It also updates the version of matplotlib in the docker/singularity versions of the tutorial and fixes a broken `sed` command in the docker version. @bgunnar5 let me know if you have questions or want me to make changes. Thanks for reviewing!


bgunnar5:
For additional context, Jane and I have discussed adding an additional setup tutorial for external users outside of LC. For this setup Jane is working on creating and packaging up a VM with Merlin/Singularity installed that users could then grab and start up. This VM would utilize the `merlin server` command which currently defaults to using Singularity; therefore, we'll default this part of the tutorial to use Singularity as well.

PR #468, opened 2024-02-21
feature/retry_priority

bgunnar5:
Added a new `RETRY` priority as the highest priority for Celery tasks so that they don't get stuck waiting in the queue with other `merlin_step` tasks that have yet to run. This was discussed in a [teams thread](https://teams.microsoft.com/l/message/19:8efc6658663c46aca4bfca4f11df7eb3@thread.skype/1708467433743?tenantId=a722dec9-ae4e-4ae3-9d75-fd66e2680a63&groupId=7c7f88bc-2485-460f-90e2-4f6880b5ed3e&parentMessageId=1708467433743&teamName=Merlin%20support&channelName=General&createdTime=1708467433743).


lucpeterson (approved):
I think we don’t intend to add freeze to pip in this PR

bgunnar5:
I initially added the call to `pip freeze` so I could debug an issue with a dependency in the CI and forgot it was in here. We might want to leave it so we can easily see the versions of our requirements that the GitHub CI is using in case anything else goes awry. It doesn't modify the functionality of the CI at all, just adds additional logs.

bgunnar5:
Or we can wait for another PR/scrap it entirely. I don't mind too much whichever way we go here.

merlin/config/utils.py
```diff
@@ -53,26 +54,31 @@ def is_redis_broker(broker: str) -> bool:
...
 
+def determine_priority_map(broker_name: str) -> Dict[Priority, int]:
+    """
+    Returns the priority mapping for the given broker name.
+
+    :param broker_name: The name of the broker that we need the priority map for
+    :returns: The priority map associated with `broker_name`
+    """
+    if is_rabbit_broker(broker_name):
+        return {Priority.LOW: 1, Priority.HIGH: 9, Priority.RETRY: 10}
```
koning, line 65: Do you want to add a MID kw here, since it is one of the options?

bgunnar5: I probably should instead of having it just default to Priority of 5. It would be easier to see what's going on this way.

merlin/config/utils.py
```diff
@@ -81,4 +81,4 @@ def get_priority(priority: Priority) -> int:
         raise ValueError(f"Invalid priority: {priority}")
 
     priority_map = determine_priority_map(CONFIG.broker.name.lower())
-    return priority_map.get(priority, 5)  # Default to 5 for MID or unknown priorities
+    return priority_map.get(priority, 5)  # Default to 5 for unknown priorities
```
koning, line 84: Might as well use it `return priority_map.get(priority, MID)`

bgunnar5: Since `Priority.MID` is an enum then I can't return it here since that will crash Celery.

koning: Yes, I changed it to the result of the enum selection.

koning (approved):

Issue #469, opened 2024-03-14
[BUG]  new “merlin status” on old workflows 


lucpeterson:
## Bug Report

**Describe the bug**
It looks like if you run the new “status” command on a workflow that was run with old merlin, it will just hang as it searches for the status files. 

**To Reproduce**
1. Run a workflow with an older version of merlin
2. Call merlin status on that with the newer version

**Expected behavior**
Some kind of warning, a direction to use the old command, and a graceful exit

bgunnar5:
So that I can keep track of them all, I'm adding on to this thread with a few additional bugs with the status command that I've encountered:

1. When a step is only using one sample, the status file is still "condensed" but then the file is just removed so you can't view the status of that step at all
2. The worker for a step is expected to be just a single worker, however several workers can process a single step (think restarting a step or multiple workers processing the same step). When multiple workers process a step, the statuses cannot be condensed properly and can cause a workflow to fail
3. For dry runs, the name has a typo when trying to render the progress bar

bgunnar5:
@lucpeterson I'm looking into the bug you mentioned now but I'm unable to reproduce. When I run a study with merlin v1.11.1 and then try to check the status with merlin v1.12.0 it just shows empty progress bars:
![image](https://github.com/LLNL/merlin/assets/49216024/a4448d38-d166-4fd8-b5e8-42c11e462124)

Which study were you running that caused this? Maybe there's a certain case where this happens but won't otherwise

lucpeterson:
This was not actually a problem

PR #471, opened 2024-04-08
bugfix/initial-status-issues

bgunnar5:
This branch should fix the problems discussed in my first comment of #469. However, I was never able to reproduce the initial bug that was posted about in that issue thread.

This branch should also allow for safer handling of the `JSONDecodeError` that users were seeing as I added a try/except for it. I think this error may have been arising due to a race condition that I failed to recognize initially. I tried to safeguard around this race condition by creating the `write_status` function. Hopefully this makes it so the `JSONDecodeError` is gone now.


merlin/study/status.py
```diff
@@ -780,45 +783,98 @@ def _remove_steps_without_statuses(self):
...
+        """
+        Apply any filters given by the --workers, --return-code, and/or --task-status arguments.
+        This function will also apply the --max-tasks limit if it was set by a user. We apply this
+        limit here so it can be done in-place; if we called apply_max_tasks_limit instead, this
+        would become a two-pass algorithm and can be really slow with lots of statuses.
+        """
+        if self.args.max_tasks is not None:
+            # Make sure the max_tasks variable is set to a reasonable number and store that value
+            if self.args.max_tasks > self.num_requested_statuses:
+                LOG.debug(
```
lucpeterson, line 813: I feel like this should maybe be a different log level. Warning, perhaps?

merlin/study/status.py
```diff
@@ -780,45 +783,98 @@ def _remove_steps_without_statuses(self):
...
+                # If a match is found, increment the number of matches found and compare against args.max_tasks limit
+                if found_a_match:
+                    matches_found += 1
+                    filtered_statuses[step_name][sub_step_workspace] = task_status_info
+                    # If we've hit the limit set by args.max_tasks, break out of the inner loop
+                    if matches_found == self.args.max_tasks:
                         break
+                else:
+                    # If our filters aren't a match for this task then delete it
+                    LOG.debug(f"No matching filter for '{sub_step_workspace}'.")
```
lucpeterson, line 858: I think this should also maybe be a Warning

merlin/study/status.py
```diff
@@ -1029,28 +1083,62 @@ def display(self, test_mode: Optional[bool] = False):
...
+        if raise_errors:
+            raise FileNotFoundError from fnf_exc
+    # Handle JSONDecode errors (this is likely due to an empty status file)
+    except json.decoder.JSONDecodeError as json_exc:
+        LOG.warning(f"JSONDecodeError raised when trying to read status from {status_filepath}")
+        if raise_errors:
+            raise json.decoder.JSONDecodeError from json_exc
 
     return statuses_read
+
```
lucpeterson, line 1129: I know it’s not pythonic, but is there a way to catch other generic errors and issue a major warning. If there’s an unhandled exception, it kills the worker silently

merlin/study/status.py
bgunnar5: yes you're right definitely a warning rather than debug

merlin/study/step.py
```diff
@@ -277,9 +266,29 @@ def _update_status_file(
...
+                # Add the current worker to the overall-step status info
+                if "workers" not in status_info[self.name]:
+                    status_info[self.name]["workers"] = [current_worker]
+                elif current_worker not in status_info[self.name]["workers"]:
+                    status_info[self.name]["workers"].append(current_worker)
+
         LOG.info(f"Writing status for {self.name} to '{status_filepath}...")
-        with open(status_filepath, "w") as status_file:
-            json.dump(status_info, status_file)
+        write_status(status_info, status_filepath, f"{self.workspace.value}/status.lock")
```
lucpeterson, line 291: I know we’re not catching errors in here, since we will want to know that something went wrong, but is there a way to do that without killing the workflow? An I/O system glitch eg that is transient could take down the workflow if not gracefully handled

merlin/study/status.py
bgunnar5: In python you can add an `except Exception` statement to catch literally anything although it's typically super frowned upon. I think for this case we can make an exception though (pun intended).

lucpeterson: Wah wah wah

lucpeterson:
Can you briefly explain the race condition and how this would fix it?

merlin/study/step.py
bgunnar5: Maybe something like I mention in my previous comment where we just catch anything and everything that could go wrong. We could instead log it with a warning, which will likely cause SOFT FAILs.

lucpeterson: I think if there’s a way we can grab bad things that happen and log them as SOFT_FAILS it’ll go a long way towards resilience

lucpeterson: As a related note (probably for a different PR) if we could do that for individual tasks, especially those in chords so we can head off chord errors, a bunch of our “frozen workflow” issues would go away (they’d become failed tasks instead, which is way better)

bgunnar5:
In `step.py` where you last commented, the status was being dumped using:

```
with open(status_filepath, "w") as status_file:
     json.dump(status_info, status_file)
```

In this case, the status is being dumped to the status file and not using a file lock object to ensure nothing else is reading/writing to it at the same time. I changed this to call:

```
write_status(status_info, status_filepath, f"{self.workspace.value}/status.lock")
```

With this change, it ensures that a file lock is being used every time the status file is being written to here.

merlin/study/step.py
bgunnar5: From a high level that sounds super do-able.

lucpeterson: I think I know how to do this (thanks Google Gemini). We can define an errback to the chord that gets called. 

See: https://docs.celeryq.dev/en/stable/userguide/canvas.html#chords

lucpeterson: They have an example that just prints the error that was called. We can also set task_allow_error_cb_on_chord_header to allow for tasks in the header to fail too

bgunnar5: I remember playing with this in the past and noticing that if a chord errback is called then that will fully stop the workflow as the typical callback (in our case the `chordfinisher` task) will never get called. When the typical callback is never given then the next chord in the chain never fires off so everything just stalls.

Maybe there's a way to trigger the next chord in the chain manually, in which case we could do that at the end of the chord errback.

lucpeterson (approved):
This looks good, right?

bgunnar5:
@lucpeterson just added two more things to this:
1. A variable in status constants that's used in merlin v1.12.0 but won't be used in v1.12.1 when this is released. This was added in case someone uses `merlin status` from v1.12.1 on a study that was ran in v1.12.0.
2. An additional if clause to check for a missing status file while condensing. This should help keep the workflow from crashing if there's a problem with condensing the status files.

Once I get the merge conflicts resolved this should be good to go

Issue #474, opened 2024-04-30
'openfoam_wf_singularity' not found


xorJane:
## Bug Report

**Description**
In #463, I introduced a bug such that `merlin example openfoam_wf_singularity` generates the error message "Example 'openfoam_wf_singularity' not found."

This occurs because of an unmet constraint on line 113 of `merlin/examples/generator.py`:

```
         if name == spec:
```

which requires that a workflow example named `foo` have an associated yaml file named `foo.yaml`. When I edited the tutorial in #463, I incorrectly named the yaml for `openfoam_wf_singularity`.

**Behavior**

```
$ merlin example openfoam_wf_singularity


       *
   *~~~~~
  *~~*~~~*      __  __           _ _
 /   ~~~~~     |  \/  |         | (_)
     ~~~~~     | \  / | ___ _ __| |_ _ __
    ~~~~~*     | |\/| |/ _ \ '__| | | '_ \
   *~~~~~~~    | |  | |  __/ |  | | | | | |
  ~~~~~~~~~~   |_|  |_|\___|_|  |_|_|_| |_|
 *~~~~~~~~~~~
   ~~~*~~~*    Machine Learning for HPC Workflows



[2024-04-29 18:02:54: ERROR] Example 'openfoam_wf_singularity' not found.
```

I'm about to submit a fix but wanted to document this. Apologies!

PR #475, opened 2024-04-30
Fix filenames for OpenFoam tutorial

xorJane:
This PR fixes a bug that I introduced in #463, as documented in Issue #474.

Currently on `develop`, running `merlin example openfoam_wf_singularity` leads to the error "'openfoam_wf_singularity' not found" because of an incorrect filename. In this PR, I update the incorrect filename and all references to said file found in the tutorial.


bgunnar5 (approved):
Approved.

This comment is just for documentation: I'm creating a huge branch full of tests and in that branch this error is tested against so we can ignore adding a test for this as of now.

xorJane:
Thank you!

PR #476, opened 2024-05-08
bugfix/deep-merge-existing-keys

bgunnar5:
This branch adds the ability to customize how merge conflicts are handled in the `dict_deep_merge` function.

This branch also adds new functionality to the test suite by providing the ability to define module-specific pytest fixtures. This will help keep the pytest fixtures more modular going forward.

There are several new tests for reading/writing the status files and also new tests for this conflict handler functionality. There are also new tests for the `dict_deep_merge` function.


merlin/utils.py
```diff
@@ -559,33 +557,81 @@ def needs_merlin_expansion(
...
+    :param `conflict_handler`: An optional function to handle conflicts between values at the same key.
+                               The function should return the value to be used in the merged dictionary.
+                               The default behavior without this argument is to log a warning.
     """
+
+    # Check to make sure we have valid dict_a and dict_b input
+    if not _both_inputs_are_dicts(dict_a, dict_b):
+        return
+
+    # # Check to make sure we have valid dict_a and dict_b input
```
lucpeterson, line 602: Unnecessary commented code

merlin/utils.py
```diff
@@ -559,33 +557,81 @@ def needs_merlin_expansion(
...
+def _both_inputs_are_dicts(dict_a: dict, dict_b: dict):
+    """
+    Helper function for `dict_deep_merge` to ensure both entries are dictionaries.
+
+    :param `dict_a`: A dict that we'll merge dict_b into
+    :param `dict_b`: A dict that we want to merge into dict_a
+    """
+    dict_a_is_dict = isinstance(dict_a, dict)
+    dict_b_is_dict = isinstance(dict_b, dict)
+    if not dict_a_is_dict or not dict_b_is_dict:
```
lucpeterson, line 569: All of this could be simplified, like msgs = [‘name is not a dict.’] for name in [A,B] if name not dict
error = ‘ ‘.join(msgs)

that syntax is wrong but you get the idea

lucpeterson (commented):
This seems pretty good. The tests do as planned?

bgunnar5:
@lucpeterson I certainly hope they do, they took forever to put together lol

I might add an integration test that starts with a partially completed workspace, then calls `merlin restart` on it just to check that it finished without erroring out.

lucpeterson:
The tests are throwing some warnings. Unclear if it’s from us or celery tho. Just FYI. They are still passing

lucpeterson (approved):

bgunnar5:
@lucpeterson I think the python 3.7 test is throwing warnings since it has an older version of kombu/celery. I think it's time we drop support for 3.7 considering it's already reached EOL. I know 3.8 reaches EOL this year as well but we should probably maintain support for it so long as LC keeps it on systems.

PR #480, opened 2024-05-16
bugfix/status_nested_workspace

bgunnar5:
This branch fixes two issues with the status command:
1. It corrects a mistake I had made when creating the merge rules. Instead of using the initial value and logging a warning for merge conflicts involving status and return code, this should instead use the newer value and log a debug statement.
2. It adds a check for nested workspaces when reading in status files that way we don't read any status files from "separate" workflows (these workflows may be nested workflows within a step, or even iterative workflows).


merlin/study/status.py
```diff
@@ -339,12 +339,37 @@ def num_requested_statuses(self):
...
+    def move_to_next_dir(self, timestamp_dirs: List[str], root: str) -> bool:
+        """
+        Look for nested workspaces and skip them.
+
+        :param timestamp_dirs: A list of nested workspaces
+        :param root: The path to the directory we need to check
+        :returns: True if we should skip this directory. False otherwise.
+        """
+        timestamp_regex = r"\d{8}-\d{6}$"
+        curr_dir = root.split("/")[-1]
```
lucpeterson, line 351: Use os.separator and other operations instead

lucpeterson:
Suggest we add some tests, especially for nested workflows logic

bgunnar5:
I'll add tests for nested workflow functionality next week and then ping you again

bgunnar5:
@lucpeterson this should be good to merge now

lucpeterson (approved):

PR #481, opened 2024-05-24
bugfix/celery-chord-error

bgunnar5:
Added a patch to Celery's result backend so that ChordErrors stop getting raised and breaking workflows. I've also added a new step return code that's mainly for testing.


bgunnar5:
Not sure what's causing python 3.7 to break when it runs the new chord error test... Maybe this is a good time to discuss whether we should drop support for it or not. It is already at EOL and has been for some time now. Thoughts?

tests/integration/test_specs/chord_err.yaml
```diff
@@ -0,0 +1,37 @@
...
+  description: Process samples. Purposefully try to trigger the chord err
+  run:
+    cmd: |
+        if [ $(SAMPLE) == "RAISE" ];
+        then
+          exit $(MERLIN_RAISE_ERROR)
+        else
+          echo "Success for sample $(SAMPLE)"
+        fi
+- name: step_2
```
lucpeterson, line 20: I think it’d be good to add another step as a dependency on step 1 but before the join step so we test chords of chains (like have 3 steps). We should also have a test with parameters and samples together

```
1-2
      \
        3
      /
1-2
```

lucpeterson:
> Not sure what's causing python 3.7 to break when it runs the new chord error test... Maybe this is a good time to discuss whether we should drop support for it or not. It is already at EOL and has been for some time now. Thoughts?

Yeah it’s hard to tell with the existing logs. Can we reproduce somehow or add verbosity?

tests/integration/test_specs/chord_err.yaml
bgunnar5: maybe step 2 can utilize parameters and samples together and then step 3 can join everything like step 2 is currently doing? I'll work on that

bgunnar5:
Just downloaded python 3.7 from source now to try to reproduce. The verbosity should already be turned on

bgunnar5:
image
Ok now I'm really confused on why the test isn't passing. I just tried with python 3.7 using 2 different sleep times for the test and it's working just fine.

lucpeterson:
Maybe add a command to the test that prints the tree of files? I wonder if it’s something about the container and maybe running out of space?

bgunnar5:
So looks like a higher sleep time is allowing it to progress further in the processing of the workflow for the new test. Not sure what is taking it so long to process the workflow compared to other python versions. I'll probably just modify the sleep time based on the python version for now.

bgunnar5:
@lucpeterson this should be ready to go now

tests/integration/definitions.py
```diff
@@ -832,7 +851,8 @@ def define_tests():  # pylint: disable=R0914,R0915
         "check chord error continues wf": {
             "cmds": [
                 f"{workers} {chord_err_wf} --vars OUTPUT_PATH=./{OUTPUT_DIR}",
-                f"{run} {chord_err_wf} --vars OUTPUT_PATH=./{OUTPUT_DIR}; sleep 10",
+                f"""{run} {chord_err_wf} --vars OUTPUT_PATH=./{OUTPUT_DIR};
+                sleep {10 if is_python_version_above_37() else 30}; tree {OUTPUT_DIR}""",
```
lucpeterson, line 855: Seems kinda hacky. Why not just up the sleep to 30 for all of them?

lucpeterson:
I’m wondering what happens if this is run in —local mode. I’m guessing the chord error still happens in that case, since this is a patch just for redis, right?

bgunnar5:
I just tried a run in local mode and it went all the way through the workflow

lucpeterson:
> I just tried a run in local mode and it went all the way through the workflow

If you remove the “fix” does the test fail? Eg j in wonder if there’s a test that’s something like

merlin —no-patch run —local chord_error.yaml -> must fail
merlin run —local chord_error.yaml —> success

lucpeterson:
> > I just tried a run in local mode and it went all the way through the workflow
> 
> If you remove the “fix” does the test fail? Eg j in wonder if there’s a test that’s something like
> 
> merlin —no-patch run —local chord_error.yaml -> must fail merlin run —local chord_error.yaml —> success

And if —local is a sufficient test we could get rid of the workers and sleep problem and add this to a local test instead of a distributed test

bgunnar5:
So it looks like the local run will pass with or without the patch. However, the distributed run does not so I think we have to keep this sleep business embedded in the test for now. I have another branch that's working on updating the test suite so maybe once I work on that more we can use some of those changes to resolve this.

PR #484, opened 2024-06-05
bugfix/flux-nodes

bgunnar5:
When flux was the native scheduler, it was not identifying the total number of nodes available in the allocation. This branch fixes that issue.

Also, this branch modifies the json schema to allow for variables to be used for the `nodes` setting in the worker definition and batch block.


merlin/study/batch.py
```diff
@@ -134,6 +134,14 @@ def get_node_count(default=1):
         the environment cannot be determined.
     :param returns: (int) The number of nodes to use.
     """
+
+    # If flux is the scheduler, we can get the size of the allocation with this
+    try:
+        get_size_proc = subprocess.run("flux getattr size", shell=True, capture_output=True, text=True)
```
lucpeterson, line 140: @koning  do we need to put a flux version check on this call to find the number of nodes?

koning: I think that was standard from the start, but I'll check.

koning: Looks like that is relatively new actually, So a version check is required.

PR #487, opened 2024-06-10
bugfix/flux-nodes-prior-versions

bgunnar5:
Added a version check for flux when getting node count of an allocation. If the version is below 0.17.0, we're not supporting it.

### DIFF
--- a/merlin/study/batch.py
+++ b/merlin/study/batch.py
@@ -40,7 +40,7 @@ import os
 import subprocess
 from typing import Dict, Optional, Union
 
-from merlin.utils import convert_timestring, get_flux_alloc, get_yaml_var
+from merlin.utils import convert_timestring, get_flux_alloc, get_flux_version, get_yaml_var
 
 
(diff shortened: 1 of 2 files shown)


merlin/study/batch.py
```diff
@@ -135,6 +135,11 @@ def get_node_count(default=1):
     :param returns: (int) The number of nodes to use.
     """
 
+    # Flux version check
+    flux_ver = get_flux_version(parsed_batch["flux exe"], no_errors=True)
+    if int(flux_ver.split(".")[1]) < 17:
```
lucpeterson, line 140: Need a check to make sure versions 1+.X.X are ok
Like major, minor, _ = flux_ver.split(‘.’)
If major < 1 and minor < 17 then error

bgunnar5: ahh yes good catch let me go fix that

lucpeterson (approved):

PR #490, opened 2024-09-13
Adding Several New Unit Tests

bgunnar5:
This branch adds _a ton_ of new unit tests for Merlin. It adds unit tests for files in:
- merlin/common
- merlin/config
- merlin/examples
- merlin/server

I plan on adding more but the amount of changes in this branch is starting to get extremely large so I think we should start the review process now.

This branch also adds in context managers to help manage the Redis server and Celery workers that will be needed for integration tests in the future. The only test file that currently uses these context managers is the `test_celeryadapter.py` file which I've moved to the integration test suite as it's more of an integration test file anyways.

As I've been writing these tests I've been cleaning up code when necessary so there are a few files that I've made changes to in the codebase as well. See the CHANGELOG for a description of these changes.

It may also be helpful to read through the README.md file of the tests folder to understand the new testing architecture I'm going for here (and let me know if it doesn't make sense so I can clear it up).


merlin/server/server_commands.py
```diff
@@ -243,6 +272,28 @@ def start_server() -> bool:  # pylint: disable=R0911
 
     LOG.info(f"Server started with PID {str(process.pid)}.")
     LOG.info(f'Merlin server operating on "{redis_out["hostname"]}" and port "{redis_out["port"]}".')
+    return True
+
+
+def start_server() -> bool:  # pylint: disable=R0911
+    """
+    Start a merlin server container using singularity.
```
lucpeterson, line 280: If we expand to things other than singularity in the future, is this the hook?

lucpeterson (commented):
Woah this is a big one. Yay testing

lucpeterson:
Have you really been working on this for a year?!?

bgunnar5:
@lucpeterson Yeah I've been working on this on and off when I got some free time since last December ish :)

merlin/main.py
```diff
@@ -425,6 +425,7 @@ def process_server(args: Namespace):
     Route to the correct function based on the command
     given via the CLI
     """
+    os.environ["LC_ALL"] = "C"  # Necessary for Redis to configure LOCALE
```
koning, line 428: You may want to check this before setting, I believe there are other viable values.

merlin/common/security/encrypt.py
```diff
@@ -52,11 +52,10 @@ def _get_key_path():
...
 
-    try:
-        key_filepath = os.path.abspath(os.path.expanduser(key_filepath))
-    except KeyError as e:
-        raise ValueError("Error! No password provided for RabbitMQ") from e
-    return key_filepath
+    if key_filepath is None:
+        raise ValueError("Error! No password provided for RabbitMQ")
+
+    return os.path.abspath(os.path.expanduser(key_filepath))
```
koning, line 58: Is an exception this may cause checked elsewhere?

merlin/main.py
bgunnar5: Maybe I should add a check that if it's set already we won't change it?

merlin/common/security/encrypt.py
bgunnar5: I don't think it's possible for `os.path.expanduser()` or `os.path.abspath()` to raise a KeyError so that's why I removed that try/except. The errors that are typically raised from these calls are `TypeError` and `OSError` in which case we probably want the exception raised anyways.

bgunnar5:
Thanks for looking at this @koning

merlin/common/security/encrypt.py
koning: OK

merlin/server/server_commands.py
```diff
@@ -243,6 +272,28 @@ def start_server() -> bool:  # pylint: disable=R0911
...
+    """
+    if not check_for_not_running_server():
+        return False
+
+    server_config = pull_server_config()
+    if not server_config:
+        LOG.error('Try to run "merlin server init" again to reinitialize values.')
+        return False
+
+    process = start_redis_container(server_config)
```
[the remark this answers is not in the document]
bgunnar5: Ok good call. I think I can just change this to `start_container` instead. I'm looking at the logic of `start_redis_container` and it still seems generalized enough that this would be acceptable.

ryannova (approved):

koning (approved):

PR #494, opened 2025-02-19
feature/monitor-auto-restart

bgunnar5:
This PR adds in the monitor auto-restart functionality that we've been discussing @lucpeterson @ryannova @rydtran . It lays the foundation for future changes I'd like to make in Merlin involving database integration and a more OOP approach to handling backend/broker connections.

No tests have been written yet as an alpha was requested ASAP. I can add documentation and tests in the days to come but for now it's probably important to get Bogdan and Ryan Tran using this and providing feedback. If we don't want to create a full alpha release just yet, maybe we can have them just use this branch for now until I get all tests and documentation finished. I'll leave that up for discussion.


bgunnar5:
Tests are failing due to Python 3.7 and the linter now using Python 3.13. Maestro is dropping support for 3.7 so I think it's time Merlin does as well. I will work on this more tomorrow.

ryanadtran:
I will go build your branch locally and test it.

bgunnar5:
@ryanadtran Ok, I'm in the process of writing some additional documentation too. For now, just know that you may want to run `merlin database delete all-runs` prior to submitting a new run. This just clears the database of existing runs so the monitor will start fresh each time.

lucpeterson (approved):
The new test seems to be working. The errors are on the jackalope server. Once those changes are incorporated I’m good with thjs

merlin/backends/redis/redis_base_store.py
```diff
@@ -0,0 +1,243 @@
...
+from redis import Redis
+
+from merlin.backends.redis.redis_utils import create_data_class_entry, deserialize_data_class, update_data_class_entry
+from merlin.db_scripts.data_models import LogicalWorkerModel, PhysicalWorkerModel, RunModel, StudyModel
+from merlin.exceptions import RunNotFoundError, StudyNotFoundError, WorkerNotFoundError
+
+
+LOG = logging.getLogger(__name__)
+
+T = TypeVar("T")
```
lucpeterson, line 25: I don't understand these variable names

bgunnar5: These were a ChatGPT addition when I asked it to help with type hinting so I'm also not 100% on this. I'll go get them changed

lucpeterson (approved):

merlin/backends/redis/redis_base_store.py
bgunnar5: I asked GPT about this and it says:

> The variable `T` is a type variable that represents a generic type parameter. The name "T" follows a widely adopted naming convention in programming:
> 
> - T stands for "Type" - it's the most common single-letter name for a generic type parameter
> - This convention comes from mathematics and has been adopted across many programming languages (Java, C#, TypeScript, Python, etc.)
> - Other common generic type variable names include `U`, `V` for additional type parameters, `K` for "Key", `V` for "Value" in key-value contexts
> 
> In your code, `T` allows `RedisStoreBase` to be generic over any type. For example:
> 
> - `RedisStoreBase[str]` - a Redis store that works with strings
> - `RedisStoreBase[User]` - a Redis store that works with User objects
> - `RedisStoreBase[dict]` - a Redis store that works with dictionaries
> 
> The `T` acts as a placeholder that gets replaced with the actual type when the class is used. It's essentially saying "this class can work with some type T, and whatever T is will be consistent throughout the class definition."
> Using single letters like `T` keeps the code concise while maintaining the established convention that developers expect to see.

Since it seems to be widely accepted, this is probably fine to keep.

PR #501, opened 2025-04-23
Requirements fixes

bgunnar5:
The DeepDiff library recently released version 5.4.2 which is breaking the unit tests for python 3.8. This branch resolves that issue.

Additionally, I noticed some requirements were specific to python 3.7 and below. Since support for 3.7 and below is dropped, these requirements have been removed. That goes for unnecessary try/except import statements as well.

### DIFF
--- a/merlin/utils.py
+++ b/merlin/utils.py
@@ -34,6 +34,7 @@ Module for project-wide utility functions.
 import getpass
 import logging
 import os
+import pickle
 import re
 import socket
 import subprocess
@@ -49,12 +50,6 @@ import pkg_resources
 import psutil
 import yaml
 from tabulate import tabulate
-
-
-try:
-    import cPickle as pickle
-except ImportError:
-    import pickle
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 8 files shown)


merlin/utils.py
```diff
@@ -51,12 +52,6 @@
 from tabulate import tabulate
 
 
-try:
```
doutriaux1, line 54: isn't cPickle better? Or is it just historical?

requirements/dev.txt
```diff
@@ -12,5 +12,6 @@ sphinx>=2.0.0
 alabaster
 johnnydep
 deepdiff
+orderly-set==5.3.0; python_version == '3.8'
```
doutriaux1, line 15: just for python 3.8 or greater than 3.8?

lucpeterson (approved):

requirements/dev.txt
bgunnar5: Just 3.8. See [this comment on orderly-set's PR that resolves this](https://github.com/seperman/orderly-set/pull/11#issuecomment-2802739146).

PR #503, opened 2025-04-29
fix an issue with docs build on python 3.9

bgunnar5:
Lina was telling me that the docs build was failing for the WEAVE CI. This happens because the WEAVE CI is using Python 3.9 and `PosixPath` does not yet have support for subscripting. The solution was to just convert `PosixPath` strings to built-in strings where this is breaking.

I also fixed 2 warnings that the griffe library was throwing at me when building the documentation.

UPDATE: I had to add a fix for pip in the CI since their latest update caused vendor package corruption.

### DIFF
--- a/.github/workflows/push-pr_workflow.yml
+++ b/.github/workflows/push-pr_workflow.yml
@@ -241,12 +241,23 @@ jobs:
 
     - name: Reinstall pip to avoid vendored package corruption
       run: |
-        curl -sS https://bootstrap.pypa.io/get-pip.py -o get-pip.py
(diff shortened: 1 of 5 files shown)


doutriaux1:
@bgunnar5 tests fail

doutriaux1:
Otherwise it's an easy approval

bgunnar5:
@doutriaux1 I'm so confused on what's going on with these tests. It looks like they're failing due to an issue with pip. However, yesterday I re-ran the unit test jobs and those just ended up passing on the second time around. When I re-run the local tests and integration tests they just continue to fail. Nothing I've changed in this PR should be breaking any tests... I'll keep looking

PR #509, opened 2025-06-09
feature/local-sqlite

bgunnar5:
This PR adds in SQLite support for local runs. This should hopefully fix the broken local tests in the monitor auto restart branch.

This PR also refactors the existing backend functionality slightly so that shared logic between Redis and SQLite is now moved up a level to the `ResultsBackend` class. Additionally, there is now a base class for database stores.


docs/user_guide/configuration/containerized_server.md
```diff
@@ -124,19 +118,21 @@ If your servers are running and set up properly, this should output something si
...
     -------------------------
 
-    config_file        | /path/to/app.yaml
+    config_file        | merlin_server/app.yaml
     is_debug           | False
-    merlin_home        | /path/to/.merlin
+    merlin_home        | ~/.merlin
     merlin_home_exists | True
-    broker server      | redis://default:******@127.0.0.1:6379/0
+    broker server      | redis://default:******@127.0.0.1:4000/0
```
doutriaux1, line 133: is the port change on purpose?

docs/user_guide/configuration/containerized_server.md
```diff
@@ -487,7 +496,7 @@ Below are the full scripts:
             try:
                 contents = yaml.load(yaml_file, yaml.FullLoader)
             except AttributeError:
-                LOG.warning(
+                print(
```
doutriaux1, line 499: print! we want log don't we?

merlin/backends/results_backend.py
```diff
@@ -74,58 +94,106 @@ def get_version(self) -> str:
...
                 - `study`
                 - `run`
                 - `logical_worker`
                 - `physical_worker`
 
         Returns:
-            BaseDataModel: The object retrieved from the specified store.
+            The object retrieved from the specified store.
 
         Raises:
```
doutriaux1, line 183: are you still raising anything?

doutriaux1 (approved):
@bgunnar5 thnaks this llook good. Did you try to scale up a bit the sqlite implementation? It can be problematic when too many instances are trying to access a sqlite file at the same time.

lucpeterson (approved):
Once checks pass I approve

docs/user_guide/configuration/containerized_server.md
bgunnar5: ah I thought I fixed this! I'll go change it again

merlin/backends/results_backend.py
bgunnar5: No not anymore, good catch.

doutriaux1:
> @doutriaux1 The sqlite backend is only intended to be used for local runs which are not run concurrently since there are no workers

@bgunnar5 and you really think that's going to stop users? 😆  Just be ready to feed questions 😉

PR #510, opened 2025-06-10
Update copyright

bgunnar5:
Copyright changes to make it easier to handle updating version and year. I added a new check in the makefile to ensure every file has a copyright header as a bunch of them were missing one. I also updated previous make commands that touched the copyright headers as they don't need to anymore.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,8 @@
 MIT License
 
-Copyright (c) 2023 Lawrence Livermore National Laboratory
+See the COPYRIGHT file for ownership and authorship information.
+
+For details, see https://github.com/LLNL/merlin.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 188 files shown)


LICENSE
```diff
@@ -1,6 +1,8 @@
 MIT License
 
-Copyright (c) 2023 Lawrence Livermore National Laboratory
+See the COPYRIGHT file for ownership and authorship information.
```
doutriaux1, line 3: Shouldn't you put that in every file it would be easier if we update the copyright in the future?

doutriaux1 (approved):
Thanks!

lucpeterson (approved):

LICENSE
bgunnar5: I'm confused on what you mean here. Every file has the following statement in a copyright header: "See top-level LICENSE and COPYRIGHT files for dates and other details." Are you talking about using the exact wording listed above?

doutriaux1: I meant wherever you have the full copyright text, why not point to the file as you do here

PR #512, opened 2025-06-11
Version/1.13.0b1

bgunnar5:
The long awaited beta release of the monitor auto-restart command! 😄


bgunnar5:
@doutriaux1 @lucpeterson The code scanning results are complaining about security vulnerabilities. I've gone through each one and generated a suggested fix using copilot. The log change ones seem easy enough to implement but the others (https://github.com/LLNL/merlin/security/code-scanning/16, https://github.com/LLNL/merlin/security/code-scanning/17, and https://github.com/LLNL/merlin/security/code-scanning/25) are more interesting. Do you think we should implement the changes that copilot suggested for these (or at least something similar)?

lucpeterson:
I think the security fixes should be addressed, but this seems like another PR into develop that needs testing. I’m particularly concerned about what doc changes need to happen re the password file and how to configure launchit

bgunnar5:
@lucpeterson after spending hours trying to research and implement this, turns out Redis _only_ accepts plaintext in the redis.conf file. Therefore, even if we encrypt the password with Fernet, Redis will just use the encrypted string as the password. I don't think there's any way around this so those last three CodeQL issues are unable to be resolved.

doutriaux1 (approved):

PR #515, opened 2025-06-26
refactor/main-module

bgunnar5:
While I was dealing with the monitor changes I got really tired of having to ctrl+f search for anything in the main.py module so I decided to refactor it into smaller pieces. I think this will make it much easier to track what's going on with different commands going forward.

No user-facing changes here; only backend changes and new tests.


lucpeterson (approved):
Nice!

PR #520, opened 2025-07-09
Marcus hsieh/task server

MarcusHsieh:
# Pluggable Task Distribution System
> **GOAL:** TaskServerInterface system for Merlin that enables pluggable task distribution backends beyond Celery. A unified interface for  task submission, coordination, and management (whilst maintaining backward compatibility).

# Changes

## Core
- [x] **TaskServerInterface**: Abstract base class (Unified task server operations)
- [x] **CeleryTaskServer**: Celery implementation
- [ ] **KafkaTaskServer**: Kafka implementation
- [x] **TaskServerFactory**: Plugin discovery and server instantiation system
- [x] **Dependency Coordination**: Backend coordination system using TaskDependency objects
- [x] **Abstraction**: Workflow submission uses TaskServerInterface methods (Replaced hardcoded Celery implementation)

## Integration
- [x] **MerlinStudy**: Task server integration and execution orchestration
- [x] **DAG Analysis**: Dependency pattern detection for coordination
- [x] **Workflow Execution**: TaskServerInterface task submission
- [x] **Error Handling**: Fallback to original implementation

## Testing
- [x] Unit tests
- [x] Mock implementations for testing framework integration
- [x] HPC environment validation

_Cont. WIP_

### DIFF
--- a/merlin/task_servers/task_server_factory.py
+++ b/merlin/task_servers/task_server_factory.py
@@ -1,0 +1,258 @@
+##############################################################################
+# Copyright (c) Lawrence Livermore National Security, LLC and other Merlin
+# Project developers. See top-level LICENSE and COPYRIGHT files for dates and
+# other details. No copyright assignment is required to contribute to Merlin.
+##############################################################################
+
+"""
+Task server factory for selecting and instantiating task servers in Merlin.
+
+This module defines the `TaskServerFactory` class, which serves as an abstraction
+layer for managing available task server implementations. It supports dynamic selection
+and instantiation of task server handlers such as Celery or (TODO) Kafka, based on user input
+or system configuration.
+
+The factory maintains mappings of task server names and aliases, and raises a clear error
+if an unsupported task server is requested.
+"""
+
+import logging
+from typing import Dict, List
+
+from merlin.task_servers.task_server_interface import TaskServerInterface
+from merlin.exceptions import MerlinInvalidTaskServerError
+
+
+LOG = logging.getLogger(__name__)
+
+
+class TaskServerFactory:
+    """
+    Factory class for managing and instantiating supported Merlin task servers.
+
+    This class maintains a registry of available task server implementations (e.g., Celery, Kafka)
+    and provides a unified interface for creating task server instances. Supports plugin discovery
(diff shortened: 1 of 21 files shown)


merlin/task_servers/task_server_factory.py
```diff
@@ -0,0 +1,258 @@
...
+        return self.list_available()
+
+    def get_task_server(self, task_server: str, config: Dict = None) -> TaskServerInterface:
+        """Legacy method - use create() instead.""" 
+        return self.create(task_server, config)
+
+    def register_task_server(self, name: str, task_server_class: TaskServerInterface,
+                           aliases: List[str] = None) -> None:
+        """Legacy method - use register() instead."""
+        return self.register(name, task_server_class, aliases)
```
bgunnar5, line 254: this is a new class so I'm not sure if we need any of these methods

merlin/task_servers/task_server_factory.py
```diff
@@ -0,0 +1,258 @@
...
+        """
+        Discover and load task server plugins from entry points and modules.
+        
+        This method attempts to find additional task server implementations
+        through Python entry points and module scanning.
+        """
+        # METHOD 1: Entry points (for pip-installable plugins)
+        try:
+            import pkg_resources
+            for entry_point in pkg_resources.iter_entry_points('merlin.task_servers'):
```
bgunnar5, line 203: This works up to python 3.11 (I think) but for 3.12+ we need to use importlib_metadata instead since they deprecated pkg_resources.

bgunnar5 (commented):
Overall great work here! This is still a draft so it obviously needs a bit more work. We should discuss some design implementations in more detail. Here's a list of topics we should talk about:

- It would probably be beneficial to create a `StudyExecutor` class and move a lot of the logic that's in the huge function of tasks.py to this class
- I'm worried about the except clauses you've added in the tasks.py file because I don't know if they'd maintain proper execution order of tasks
- We should likely remove router.py file as the task server implementations will handle this
- I think the functions in celeryadapter.py should probably be moved to the celery_server.py file eventually. Although, that may make the file too large... not sure yet.

PR #524, opened 2025-08-05
feature/filter-db-queries

bgunnar5:
This PR adds filtering to the database `get all-*` and `delete all-*` commands. This is needed for the refactor of `query-workers` that's coming soon.


bgunnar5:
TODO: add documentation

PR #528, opened 2025-08-27
refactor/query-workers

bgunnar5:
_This branch is an extension of #523 and #524 so both of these need to be merged first. Once those are merged the number of changed files should decrease dramatically._

This branch refactors the `merlin query-workers` command to utilize the new `MerlinWorker` classes. It also drastically improves the output of this command by utilizing the new `MerlinDatabase` to query worker information. The output of this command is now defaulted to use rich tables that automatically detect the width of the terminal in order to improve readability for users. There is also a new `json` format option to dump the output to a json string.

_This PR is still a draft as I need to write tests for all of the new functionality._


docs/user_guide/command_line.md
```diff
@@ -1497,7 +1497,8 @@ merlin query-workers [OPTIONS]
 | `--task_server`  | string | Task server type for which to query workers. Currently only "celery" is implemented. | "celery" |
 | `--spec` | filename | Query for the workers named in the `merlin` block of the spec file given here | None |
 | `--queues` | List[string] | Takes a space-delimited list of queues as input. This will query for workers associated with the names of the queues you provide here. | None |
-| `--workers` | List[regex] | A space-delimited list of regular expressions representing workers to query | None |
+| `--workers` | List[string] | A space-delimited list of logical worker names to query | None |
```
doutriaux1, line 1500: no more regex?

merlin/abstracts/factory.py
```diff
@@ -227,6 +213,7 @@ def _get_component_class(self, canonical_name: str, component_type: str) -> Any:
 
         return component_class
 
+    # TODO should we change 'config' to 'kwargs'?
```
doutriaux1, line 216: sounds like a good idea, but you'll have to keep config as a `kwarg` for backward compatibility

merlin/db_scripts/entities/physical_worker_entity.py
```diff
@@ -202,9 +202,18 @@ def get_pid(self) -> Optional[int]:
             The process ID for this worker or None if not set.
         """
         self.reload_data()
-        return int(self.entity_info.pid) if self.entity_info.pid else None
+        self.reload_data()
```
doutriaux1: `reload_data` twice?

merlin/db_scripts/entities/physical_worker_entity.py
```diff
@@ -294,7 +303,7 @@ def get_restart_count(self) -> int:
             The number of times that this worker has been restarted.
         """
         self.reload_data()
-        return self.entity_info.restart_count
+        return int(float(self.entity_info.restart_count))
```
doutriaux1, line 305: can't you just do `int()` ? does it contain the `.` in it?

merlin/workers/formatters/__init__.py
doutriaux1, line 1: I like this, I actually never added a doc string to my `__init__.py` files.

merlin/workers/handlers/worker_handler.py
```diff
@@ -25,14 +26,23 @@ class MerlinWorkerHandler(ABC):
...
-    def __init__(self):
-        """Initialize the worker handler."""
+    def __init__(self, merlin_db: MerlinDatabase = None):
+        """
+        Initialize the worker handler.
+
+        Args:
+            merlin_db: The database instance used for worker management or None.
+        """
+        self.merlin_db = merlin_db or MerlinDatabase()
```
doutriaux1, line 45: neat

doutriaux1 (approved):
That looks ready to go. Thanks @bgunnar5

docs/user_guide/command_line.md
bgunnar5: No, how this is querying workers behind the scenes is totally different now so I elected to just have this be a list of worker_names only for now. I'll probably go through and add regex capabilities in the future but this PR is already really large.

merlin/abstracts/factory.py
bgunnar5: This class is new to the 2.0 beta so I'm not sure if I'd need to leave it in there or not. I've warned users that code in the beta is subject to breaking changes

merlin/db_scripts/entities/physical_worker_entity.py
bgunnar5: oops good catch!

merlin/db_scripts/entities/physical_worker_entity.py
bgunnar5: Trying to remember what was going on here... I think when we reload the data it queries the backend database (redis or sqlite). This is returned as a string and will contain the `.` so we have to use float first.

merlin/workers/formatters/__init__.py
bgunnar5: It actually shows up in the API documentation which is helpful! See [here](https://merlin.readthedocs.io/en/develop-2.0/api_reference/backends/) for an example.

PR #585, opened 2025-10-07
feature/db-garbage-collection

### LINKED ISSUE #582: Garbage Collection For Database
The database needs some sort of automatic garbage collection functionality. This is necessary since the monitor will run and sometimes get stuck on a workflow that no longer exists.

bgunnar5:
Adds garbage collection functionality for the database. Additionally, this garbage collection is integrated into the `merlin monitor` command so that it's run by default. There are also some extra checks for invalid workflows that have been added to the monitor to try to make it more resilient.


merlin/db_scripts/garbage_collector.py
```diff
@@ -0,0 +1,429 @@
...
+
+        # Get all valid run IDs from the database
+        all_runs = self.merlin_db.get_all("run")
+        valid_run_ids = {run.get_id() for run in all_runs if run.get_id() not in invalid_run_ids}
+
+        all_logical_workers = self.merlin_db.get_all("logical_worker")
+        for worker in all_logical_workers:
+            worker_runs = worker.get_runs()
+            # Worker is orphaned if:
+            # - It has no runs, OR
```
lucpeterson, line 231: Are workers orphaned if we haven’t done merlin run yet? (Eg we launch workers first and they’re just chillin’ before we push work)? What happens to them in that case?

bgunnar5: Great point and one I overlooked initially. I can go change the monitor to only garbage collect studies and runs as that's really where the problems have been happening.

When I implement the worker watchdog process that you, me, and Ryan Lee had discussed in the past then we may need to think about this more. Will the worker watchdog need to call this garbage collector at all? I'm not sure as of now.

docs/user_guide/database/garbage_collection.md
```diff
@@ -0,0 +1,213 @@
...
+- Storage cleanup operations remove old run directories
+- Studies are moved or archived without updating the database
+
+The garbage collection process identifies these orphaned entries and can remove them automatically, keeping your database synchronized with the actual state of your filesystem.
+
+### How Garbage Collection Works
+
+The garbage collection process follows a cascading cleanup approach:
+
+1. **Identifies orphaned runs**: Scans all run entries in the database and checks if their workspace directories still exist
```
jwhite242, line 24: I'm not to the code implementing this yet, so may see the answer there, but what about workspaces which do exist, but are simply not visible by the process doing the garbage collection given the machine it is running on?

docs/user_guide/database/garbage_collection.md
```diff
@@ -0,0 +1,213 @@
...
+    ============================================================
+
+    Invalid Runs: 1
+      - /path/to/hello_samples_20251008-102348
+
+    Orphaned Logical Workers: 1
+      - hello_samples_worker (queues: [merlin]_step_1_queue, [merlin]_step_2_queue)
+
+    Orphaned Physical Workers: 1
+      - celery@hello_samples_worker.%dane13 (host: dane13)
```
jwhite242: the '%' before dane13 intentional here?

docs/user_guide/database/garbage_collection.md
```diff
@@ -0,0 +1,213 @@
...
+merlin database gc --skip-workers --skip-studies
+```
+
+## Limitations
+
+The garbage collection process:
+
+- Only checks for missing workspace directories (it does not validate workspace contents)
+- Does not clean up data in the workspace directories themselves (only database entries)
+- Requires read access to the filesystem paths referenced in run entries
```
jwhite242, line 295: Might want to call this one out more loudly given it may dictate where they can/should run this command.  Additionally, what happens if you do it on a machine that doesn't have read access to all the filesystems and thus can't see if the workspaces exist? does it clobber them anyway, assuming they're just not there?  (and read isn't just permissions, but potentially completely unavailable filesystems, likely giving different error conditions)

docs/user_guide/database/info.md
```diff
@@ -0,0 +1,151 @@
...
+    1. ID: 2f740737-a727-ea7d-6de4-17dc643183bb, Name: hello_worker, Queues: ['[merlin]_merlin']
+    2. ID: 4b0cd8f6-35a3-b484-4603-fa55eb0e7134, Name: hello_samples_worker, Queues: ['[merlin]_step_1_queue', '[merlin]_step_2_queue']
+
+Physical Workers:
+- Total: 4
+- Recent Physical Workers:
+    1. ID: 8549ed5f-83df-4922-aaac-16f676112322, Name: celery@hello_worker.%ruby9
+    2. ID: dbf0a091-6b7f-4ade-b404-6f00a104049a, Name: celery@hello_worker.%ruby11
+    3. ID: 9a6b8bec-2ede-4a8c-bb07-0778c5c5f356, Name: celery@hello_samples_worker.%ruby10
+    4. ID: bab9a016-9ae9-40ad-8a80-8e46a21e3fb7, Name: celery@hell_samples_worker.%ruby12
```
jwhite242, line 110: One more thought on this .% givne it looks to be intentional: beware that %r is an actual operator in python's old string formatting -> any downstream parsing of this output form via log scrapers or anything would have to use the new style only to be safe (the %d in %dane13 suffers similar overloading).

docs/user_guide/database/retrieving_data.md
```diff
@@ -0,0 +1,345 @@
...
+```
+
+The output includes the following fields:
+
+| Field            | Description                                                       |
+|------------------|-------------------------------------------------------------------|
+| **ID**           | A unique identifier for the study.                                |
+| **Name**         | The name assigned to the study.                                   |
+| **Runs**         | A list of associated runs, with details such as ID and workspace. |
+| **Additional Data** | Any extra metadata stored with the study.                      |
```
jwhite242, line 63: What might show up here, and is this something users have control of to inject misc info here?

docs/user_guide/database/retrieving_data.md
```diff
@@ -0,0 +1,345 @@
...
+Name: hello_samples
+Runs:
+  - ID: c735ade0-9b28-4b9e-bb46-d9429d7cf61a
+    Workspace: /path/to/hello_samples_20250508-161159
+Additional Data: {}
+```
+
+## Retrieving All Runs
+
+Using the `all-runs` option will retrieve and display each run in our database. When displayed, each run will display every field stored in a run entry.
```
jwhite242, line 89: what is a 'run'?   study, single step in a study, chains in a study's graph (multiple steps..)  The output looks like studies, which begs the question: should this be 'all-studies' instead to avoid overloading the same term many users use for single simulations in a study?

docs/user_guide/database/retrieving_data.md
```diff
@@ -0,0 +1,345 @@
...
+```
+
+The output includes the following fields:
+
+| Field                | Description                                                      |
+|----------------------|------------------------------------------------------------------|
+| **ID**               | A unique identifier for the logical worker.                      |
+| **Name**             | The name of the logical worker from the spec.                    |
+| **Runs**             | The runs utilizing this logical worker.                          |
+| **Queues**           | A list of queues that the logical worker is watching.            |
```
jwhite242: Watching feels like an odd choice given the logical workers seem more like a config, i.e. nothing's watching a queue until a physical worker is created?

docs/user_guide/database/retrieving_data.md
```diff
@@ -0,0 +1,345 @@
...
+Host: ruby10
+Restart Count: 0.0
+Additional Data: {}
+```
+
+## Retrieving Everything
+
+!!! tip "Large Databases"
+
+    If you have a large database with many entries, the `everything` option may produce substantial output. Consider using the [`merlin database info`](./info.md) command first to see how many entries exist, or use the specific entity type retrieval commands to get only what you need.
```
jwhite242: A pager might be another good suggestion -> not necessarily built into merlin directly, but mention piping this output into something like `less` which has built in searching of the stream?

docs/user_guide/monitoring/monitor_for_allocation.md
```diff
@@ -31,9 +31,20 @@ The resulting flowchart of this process can be seen below.
...
 </figure>
 
+## Database Garbage Collection
+
+When the monitor first starts, it automatically performs [garbage collection](../database/garbage_collection.md) on the database to identify and remove stale entries. This cleanup process specifically targets:
+
+- **Runs:** Removes database entries for runs whose corresponding workspace directories no longer exist on the filesystem
+- **Studies:** Removes database entries for studies that have no associated runs remaining after the run cleanup
+
+Worker entries (both logical and physical), are not afftected by this automatic garbage collection, as they may be spun up prior to a run being executed.
```
jwhite242: afftected->affected

merlin/db_scripts/garbage_collector.py
```diff
@@ -0,0 +1,429 @@
...
+            # - All its runs don't exist in the database anymore
+            if not worker_runs or all(run_id in invalid_run_ids or run_id not in valid_run_ids for run_id in worker_runs):
+                self._issues["logical_workers"].append(worker)
+
+        LOG.info(f"[GARBAGE COLLECTOR] Found {len(self._issues['logical_workers'])} orphaned logical workers.")
+        LOG.debug(f"[GARBAGE COLLECTOR] Orphaned logical workers: {self._issues['logical_workers']}")
+
+    def _check_orphaned_physical_workers(self):
+        """
+        Check for physical workers not associated with any active logical workers.
```
jwhite242, line 242: Can it happen that you don't want these cleaned up in some cases?  I.e. what about the iterative case where you might reuse the same queue/logical worker config on the next study in the chain and thus might expect some workers in an allocation you want to keep alive to be temporarily orphaned.  Is such a case handled downstream of the issue tracker mentioned in the docstring?

merlin/db_scripts/garbage_collector.py
```diff
@@ -0,0 +1,429 @@
...
+        # Get the current invalid run IDs
+        invalid_run_ids = [run.get_id() for run in self._issues["runs"]]
+
+        # Get all valid run IDs from the database
+        all_runs = self.merlin_db.runs.get_all()
+        valid_run_ids = {run.get_id() for run in all_runs if run.get_id() not in invalid_run_ids}
+
+        all_studies = self.merlin_db.studies.get_all()
+        for study in all_studies:
+            runs = study.get_runs()
```
jwhite242, line 293: so a run is a 'step' in your study?    felt more like runs were ~studies in some of the earlier documentation on the pr on the other db commands

merlin/db_scripts/garbage_collector.py
```diff
@@ -0,0 +1,429 @@
...
+
+        if not self._issues[entity_type]:
+            LOG.info(f"[GARBAGE COLLECTOR] No stale {entity_type} found.")
+            return
+
+        LOG.info(f"[GARBAGE COLLECTOR] Deleting {len(self._issues[entity_type])} {entity_type}...")
+        for entity in self._issues[entity_type]:
+            entity_id = entity.get_id()
+            try:
+                if entity_type == "studies":
```
jwhite242: Is this plural just for the log messages above? I.e. why not have it be 'study' as used in the delete call below?   an entity of type studies implies something a little different than just a singular study

merlin/db_scripts/garbage_collector.py
```diff
@@ -0,0 +1,429 @@
...
+                    self.merlin_db.delete("study", entity_id, remove_associated_runs=False)
+                else:
+                    self.merlin_db.delete(entity_singular, entity_id)
+                LOG.debug(f"[GARBAGE COLLECTOR] Deleted {entity_singular} {entity_id}")
+            # pylint complains about broad-exception but don't think we care here
```
jwhite242: I'd normally agree with pylint here and try and catch more specific things when possible if you have more info on what can happen here since catching arbitrary exceptions can make the log messages confusing when it's something completely unrelated.   Do you know what kinds of errors can be thrown in here and try capturing those leaving the generic exception + a new messages as a fall-through for truly unknown errors?

merlin/db_scripts/garbage_collector.py
```diff
@@ -0,0 +1,429 @@
...
+
+    def cleanup_runs(self):
+        """Remove runs with invalid workspaces."""
+        self._cleanup_entity("runs")
+
+    def cleanup_workers(self):
+        """
+        Remove orphaned workers (both physical and logical).
+
+        This is done in order.
```
jwhite242: What order?  Also, does the order matter if you're cleaning up both?

merlin/db_scripts/garbage_collector.py
```diff
@@ -0,0 +1,429 @@
...
+        confirmation unless force=True.
+
+        Args:
+            check_runs: Whether to check for invalid run workspaces.
+            check_workers: Whether to check for orphaned workers.
+            check_studies: Whether to check for empty studies.
+            force: If True, skip confirmation prompt (use with caution).
+
+        Raises:
+            ValueError: If no scan has been performed yet.
```
jwhite242: Doesn't look like this one raises an exception

merlin/monitor/monitor.py
```diff
@@ -59,7 +61,9 @@ class Monitor:
         restart_workflow: Restart a run of a workflow.
     """
 
-    def __init__(self, spec: MerlinSpec, sleep: int, task_server: str, no_restart: bool):
+    def __init__(
+        self, spec: MerlinSpec, sleep: int, task_server: str, no_restart: bool, auto_cleanup: bool = True
+    ):  # pylint: disable=too-many-arguments,too-many-positional-arguments
```
jwhite242: there some other config for pylint to control this?  having 5 arguments trip this seems way too low

tests/unit/cli/commands/database/test_gc_command.py
```diff
@@ -0,0 +1,432 @@
...
+
+    Args:
+        command: Instance of the `DatabaseGarbageCollectionCommand` under test.
+    """
+    parser = ArgumentParser()
+    subparsers = parser.add_subparsers(dest="subcmd", required=True)
+    command.add_parser(subparsers)
+
+    args = parser.parse_args(["gc"])
+    assert args.func == command.process_command
```
jwhite242, line 66: Is this too granular of a test?  isn't the important thing more that the operations work, not that this collectioncommand specifically sets 'process_command' as the function to do it?

jwhite242 (commented):
One more general comment: type usage in the method argument docstrings was ~inconsistent.

docs/user_guide/database/garbage_collection.md
bgunnar5: For now, yes. This was something that was added to the worker launch process prior to me even starting at the lab:

```
        # TODO: Once flux fixes their bug, change this back to %h
        # %h in Celery is short for hostname including domain name
        worker_args += f" -n {worker_name}{nhash}.%%h"
```

I've honestly forgotten about it and should probably go take a look at that on another branch.

docs/user_guide/database/garbage_collection.md
bgunnar5: This is a good point. Hmmm... I will probably need to add some sort of hostname check here. I'd be open to suggestions if you have any

docs/user_guide/database/garbage_collection.md
bgunnar5: Yeah, this ties into your [previous comment](https://github.com/LLNL/merlin/pull/585#discussion_r2427148592). I think we need to add some sort of mechanism that makes it so we are able to check which workspace exists based on the host, in which case I'll need to update these docs.

docs/user_guide/database/info.md
bgunnar5: Yeah, I'll go see if this is even still needed in another branch.

docs/user_guide/database/retrieving_data.md
bgunnar5: For now this is sort of a placeholder that isn't being used. I was thinking that there may be a case in the future where the user wants to put more metadata in the `description` block than just `name` and `description`. In that case we'd likely need a place to store that metadata as well.

After saying this, not sure why there isn't a `description` entry here... I think that was a miss on my part.

docs/user_guide/database/retrieving_data.md
bgunnar5: Right, so this would probably be a good time to get some feedback on this page in the docs: https://merlin.readthedocs.io/en/develop-2.0/user_guide/database/entities/. If the definitions of each term in the database aren't clear after this then please let me know so I can go update that page. I should mention that some of the entries listed for `RunEntity` have not actually been implemented yet since this is still in beta (e.g., `steps`, `parent`, `child`, `parameters`, `samples`)

I have plans in the future to add `StepEntity` and `TaskEntity` models in order to help with tracking status of studies. My initial thought was to have these models represent the following:

1. `StepEntity`: essentially a logical task entry; this will be exactly what the user provides in the spec file for a single step
2. `TaskEntity`: the actual task that's being ran (so parameter/sample combos, expanded LAUNCHER, linked to worker that's processing the task, etc.)

Since all of this database business is still in beta form then I'm definitely open to having a discussion on naming conventions and where things should be stored.

docs/user_guide/database/retrieving_data.md
bgunnar5: Good catch, perhaps "A list of queues that the logical worker’s physical instances will consume tasks from." would be better?

docs/user_guide/database/retrieving_data.md
bgunnar5: Good suggestion. Adding now.

merlin/db_scripts/garbage_collector.py
bgunnar5: So with the `merlin monitor`, which is what keeps allocations alive while the workflow is running, the garbage collection process is only ran one time at the startup of the monitor. When it switches to a new run in the iteration, garbage collection won't be ran so for the iterative case you shouldn't have to worry about this I don't think?

With this being said, maybe there would be a user in the future who wants to disable checking for orphaned physical workers. I already have `--skip-runs`, `--skip-workers`, and `--skip-studies` options. I could just make this more granular and add `--skip-logical-workers` and `--skip-physical-workers` (then the `--skip-workers` option just becomes shorthand for using both of these).

merlin/db_scripts/garbage_collector.py
bgunnar5: Again it may be helpful to reference this page in the docs https://merlin.readthedocs.io/en/develop-2.0/user_guide/database/entities/. So for terminology purposes:

- "study": basically just the `description` block of your spec file. Each study entry is unique by name and contains a list of "run" entries
- "run": a full execution of a "study"

In creating this I was thinking you may have the same yaml file that you run multiple times. Hence the `study.get_runs()` call here is really retrieving all executions of this study.

Open to suggestions on naming conventions as mentioned before.

merlin/db_scripts/garbage_collector.py
bgunnar5: So the plural is used as the keys in the `self._issues` attribute of the garbage collector as I felt like that made more sense to read. The database call needs a singular form of the entity type though so maybe to stay consistent with that, this `issues` tracker should also use the singular form?

merlin/db_scripts/garbage_collector.py
bgunnar5: So did some digging on this and the only errors that I expect to be raised here are `RunNotFoundError`, `StudyNotFoundError`, and `WorkerNotFoundError` for when the respective entry isn't found in the database. I'll update this to catch those instead and let other exceptions actually be raised.

merlin/db_scripts/garbage_collector.py
bgunnar5: Order being deleting physical workers first, then logical workers. This is done because the delete method of the physical worker manager looks like so:

```python
def delete(self, worker_id_or_name: str):
        """
        Delete a physical worker entity by its ID or name.

        This method will also attempt to remove the deleted physical worker's ID
        from its associated logical worker. If the logical worker cannot be found,
        a warning is logged and deletion continues.

        Args:
            worker_id_or_name (str): The ID or name of the physical worker to delete.
        """

        def cleanup_physical_worker(worker):
            logical_worker_id = worker.get_logical_worker_id()
            try:
                logical_worker = self.db.logical_workers.get(worker_id=logical_worker_id)
                logical_worker.remove_physical_worker(worker.get_id())
            except WorkerNotFoundError:
                LOG.warning(
                    f"Couldn't find logical worker with id {logical_worker_id}. Continuing with physical worker delete."
                )

        self._delete_entity(PhysicalWorkerEntity, worker_id_or_name, cleanup_fn=cleanup_physical_worker)
```

Since the cleanup function is checking for logical workers first, then in order to avoid logging a bunch of warnings I'm having physical workers be deleted prior to logical workers.

merlin/db_scripts/garbage_collector.py
bgunnar5: You're right. My first draft did but decided to remove it.

merlin/monitor/monitor.py
bgunnar5: That was my thought too. Pylint sees it as 6 arguments because of `self` and that's why it complains. I'll have to look into this. Does Maestro use pylint and control this setting?

tests/unit/cli/commands/database/test_gc_command.py
bgunnar5: This test file is specifically testing the entry point of the garbage collection command as that's what the files in the `merlin/cli/commands/` folder contain. The more granular tests that I think you're looking for are in the [`tests/unit/db_scripts/test_garbage_collector.py`](https://github.com/LLNL/merlin/pull/585/files#diff-cfec808f1b1726f045add277ed46c51383e53bba9d51054ed3ffdfb042ccd039) file. Let me know if I'm mistaken.

bgunnar5:
Thanks for the review @jwhite242. I'll go back through all of the docstrings/type hints to make sure they're accurate with the latest changes once this is done.

docs/user_guide/database/garbage_collection.md
```diff
@@ -204,10 +270,28 @@ merlin database gc --skip-runs --dry-run
 merlin database gc --skip-workers --skip-studies
 ```
 
+## Best Practices for Distributed Environments
+
+When working across multiple machines with different filesystem access:
+
+1. **Run from a machine with broad filesystem access**: For the most comprehensive cleanup, run garbage collection from a machine that has access to the most access shared file systems.
```
jwhite242: Seems like 2 and 3 are probably the most important?  I.e. those two are what you should always do (regardless of how many machines you have to use) before falling back to manual deletion/cleanup.   This one seems more like: know your filesystem topologies/mounts, with a bonus being if you have on that can actually mount everything to do fewer iterations on 2 and 3?

docs/user_guide/database/retrieving_data.md
```diff
@@ -334,7 +334,7 @@ Additional Data: {}
 
 !!! tip "Large Databases"
 
-    If you have a large database with many entries, the `everything` option may produce substantial output. Consider using the [`merlin database info`](./info.md) command first to see how many entries exist, or use the specific entity type retrieval commands to get only what you need.
+    If you have a large database with many entries, the `everything` option may produce substantial output. Consider using the [`merlin database info`](./info.md) command first to see how many entries exist, or use the specific entity type retrieval commands to get only what you need. You can also pipe the output into a pager such as less to make browsing and searching easier.
```
jwhite242: backticks around 'less' to make it clear it's code/tool

tests/unit/cli/commands/database/test_gc_command.py
jwhite242: well, was more just that it seemed like you were testing argparse itself when testing the commands functionality themselves covers this same code; little more granular than I'd expected is all.

jwhite242 (approved):
Thanks for addressing the cross-filesystem issues!

docs/user_guide/database/garbage_collection.md
bgunnar5: Good point, I'm going to re-order these.

PR #587, opened 2025-10-22
bugfix/monitor-multi-run

### LINKED ISSUE #586: Multi-Chain Monitor Functionality
When multiple chains of the same iterative workflow are launched and using the same pool of workers, the monitor is only monitoring a single chain (more specifically a single iteration in that chain) at a time. We need a way for this monitor to watch all chains.

The solution here is likely to update the while loop in the monitor's logic so that it's looping through all of the runs associated with a study rather than stopping at one run until it's complete.

bgunnar5:
When launching multiple chains of the same workflow, the monitor used to be set up to only watch a single run until it runs to completion. This PR changes that so now the monitor will loop through all runs for a study and ensure they all complete.


lucpeterson (approved):
This looks pretty good. Do we think the tests are sufficient?

bgunnar5:
@lucpeterson I think they are but then again I'm biased cause I wrote them

merlin/monitor/monitor.py
```diff
@@ -47,17 +47,25 @@ class Monitor:
...
 
+    The monitor supports tracking multiple concurrent runs of the same study, making it suitable for
+    iterative workflows and scenarios where multiple runs share the same worker pool. It continuously
+    polls the database for active runs and performs health checks on each one during every monitoring
+    cycle.
+
     Attributes:
         spec (MerlinSpec): The Merlin specification that defines the workflow.
         sleep (int): The interval (in seconds) between monitoring checks.
-        no_restart (bool): If True, the monitor will not try to restart the workflow.
```
doutriaux1, line 53: niced attention to detail here. I probably would have forgotten to update this sentence.

merlin/monitor/monitor.py
```diff
@@ -191,94 +174,159 @@ def _check_task_activity(self, run: RunEntity) -> bool:
...
-                # If no tasks are in the queues or being processed by workers and the run is not complete, we have a hanging
-                # workflow so restart it
-                if not active_tasks and not run_complete:
-                    if self.no_restart:
-                        LOG.warning(
-                            f"Monitor: Determined restart was required for '{run_workspace}' but auto-restart is disabled."
-                        )
-                    else:
-                        self.restart_workflow(run)
+        while True:
```
doutriaux1, line 255: that always scares me!

doutriaux1 (approved):

merlin/monitor/monitor.py
bgunnar5: me too, but less so in this case since the monitor is intended to stay up and running for the duration of the allocation anyways.

PR #596, opened 2025-12-02
refactor/remove-old-monitor-code

bgunnar5:
This PR removes the old v1.0 Monitor code that's no longer used. Really this is just trying to clean up the codebase a bit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to Merlin will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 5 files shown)


doutriaux1 (approved):
Completely removed? You don't want to leave something in to say "deprecated"?

bgunnar5:
@doutriaux1 This will be going into Merlin v2.0 so I think it's safe to completely remove it.

doutriaux1:
> @doutriaux1 This will be going into Merlin v2.0 so I think it's safe to completely remove it.

@bgunnar5 ok then it's fine. Are you putting a warning in develop then?

bgunnar5:
@doutriaux1 I can. The only thing that's really changing here is the backend functionality and the `--steps` option being removed. From the frontend, the user can still use `merlin monitor ${YAML}` with the exact same call.

